PR #11, opened 2018-10-23
Add generic collection support. Fixes #2

### LINKED ISSUE #2: Add generic collection support
The following syntax must be supported:

```php
class MyObject
{
    /** @var \Spatie\ValueObject\T[] */
    public $arrayOfT;
}
```

When setting this property, we must validate two things:

- The value is an array.
- This array only contains values of type `T`.

The implementation will be in `Property::assertTypeEquals`. I expect a new check to be added which detects the array syntax `[]` and will recursively validate every element in it. `assertTypeEquals` should be reused. 

`null` values should not be supported.

bhulsman:
Although you tagged it with 'good first issue', I'll admit it wasn't easy :p There is probably a better way to solve this, but I'll submit the PR anyway, maybe get some good feedback. If I'm wasting your time; sorry, but just close it, no hard feelings.

Some issues I ran into: 

1. `ValueObjectException` should get a value of the class in the collection, not the array the class is in. But that would require changes for `set` in `Property`. I added an `is_array` check to `ValueObjectException` to avoid `array to string` issues , but there should be a better way to handle this.
2. `assertTypeEquals` in `Property` will recursively walk through the values of the array to check if the type matches. I feel like this should maybe be done in `isValidType`, but how combine this with the support for multiple types?
3. Looking at `resolveTypeDefinition` I'm not sure how to handle support for multiple types. Besides that, would probably add a `isGenericCollection` property to solve issues we now have in point 1 and 2.

I would love some feedback in order to improve the PR and maybe you can specify handling multiple types with generics. Not sure if there are any use cases for it. For example FooClass or a collection of FooClass.

### DIFF
--- a/src/Property.php
+++ b/src/Property.php
@@ -107,7 +107,7 @@ class Property extends ReflectionProperty
     {
         if (is_array($value)) {
             foreach ($value as $class) {
-                if (!$this->assertTypeEquals($type, $class)) {
+                if (! $this->assertTypeEquals($type, $class)) {
                     return false;
                 }
             }
--- a/src/Property.php
+++ b/src/Property.php
@@ -67,7 +67,7 @@ class Property extends ReflectionProperty
             return;
         }
 
-        preg_match('/\@var ([\w|\\\\]+)/', $docComment, $matches);
+        preg_match('/\@var ([\w|\\\\\[\]]+)/', $docComment, $matches);
 
         if (! count($matches)) {
             return;
--- a/src/Property.php
+++ b/src/Property.php
@@ -105,14 +105,8 @@ class Property extends ReflectionProperty
 
     protected function assertTypeEquals(string $type, $value): bool
     {
-        if (is_array($value)) {
-            foreach ($value as $class) {
-                if (! $this->assertTypeEquals($type, $class)) {
-                    return false;
-                }
-            }
-
-            return true;
+        if (strpos($type, '[]') !== false) {
+            return $this->isValidGenericCollection($type, $value);
         }
 
         if ($type === 'mixed' && $value !== null) {
@@ -124,5 +118,22 @@ class Property extends ReflectionProperty
         }
 
         return gettype($value) === (self::$typeMapping[$type] ?? $type);
+    }
+
+    protected function isValidGenericCollection(string $type, $collection): bool
+    {
+        if (! is_array($collection)) {
+            return false;
+        }
+
+        $valueType = str_replace('[]', '', $type);
+
+        foreach ($collection as $value) {
+            if (! $this->assertTypeEquals($valueType, $value)) {
+                return false;
+            }
+        }
+
+        return true;
     }
 }
--- a/src/Property.php
+++ b/src/Property.php
@@ -67,7 +67,7 @@ class Property extends ReflectionProperty
             return;
         }
 
-        preg_match('/\@var ([\w|\\\\\[\]]+)/', $docComment, $matches);
+        preg_match('/\@var ((?:(?:[\w|\\\\])+(?:\[\])?)+)/', $docComment, $matches);
 
         if (! count($matches)) {
             return;
--- a/src/Property.php
+++ b/src/Property.php
@@ -105,6 +105,16 @@ class Property extends ReflectionProperty
 
     protected function assertTypeEquals(string $type, $value): bool
     {
+        if (is_array($value)) {
+            foreach ($value as $class) {
+                if (!$this->assertTypeEquals($type, $class)) {
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
         if ($type === 'mixed' && $value !== null) {
             return true;
         }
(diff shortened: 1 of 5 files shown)


brendt:
Thanks! I appreciate the fast response! Are you up for feedback? If so, I'll gladly take the time and guide you through it. If not, no hard feelings either :)

bhulsman:
> Thanks! I appreciate the fast response! Are you up for feedback? If so, I'll gladly take the time and guide you through it. If not, no hard feelings either :)

Sure, that would be great! 

There's always plenty to learn and who would turn away from some solid feedback, right? 8)

src/Property.php
```diff
@@ -105,6 +105,16 @@ protected function isValidType($value): bool
 
     protected function assertTypeEquals(string $type, $value): bool
     {
+        if (is_array($value)) {
```
brendt: With your approach, you're inferring the type "array of" based on the value. It should be the other way around: the type should be defined, and the value should follow that type. If not, a type error is thrown.

In practice you'll want to validate whether `$type` indicates an array. Say for example that `$type` is `string[]`:

```php
if (strpos('[]', $type) !== false)
{
    if (! is_array($value)) {
        return false;
    }

    $arrayType = str_replace('[]', '', $type);

    foreach ($value as $valueItem) {
        $this->assertTypeEquals($arrayType, $valueItem);
    }
}
```

(Be sure to double check whether I got the parameter ordening right in the `str` functions 😝)

Perhaps it would be good to split this functionality into a separate function, to keep `assertTypeEquals` cleaner. You can call that function in `assertTypeEquals` of course.

tests/ValueObjectTest.php
```diff
@@ -125,6 +126,35 @@ public function classes_are_supported()
         };
     }
 
+    /** @test */
+    public function generic_collections_are_supported()
+    {
+        new class(['foo' => [new DummyClass(), new DummyClass(), new DummyClass()]]) extends ValueObject {
```
brendt: One array element should suffice in this test.

tests/ValueObjectTest.php
```diff
@@ -125,6 +126,35 @@ public function classes_are_supported()
...
+            /** @var \Spatie\ValueObject\Tests\TestClasses\DummyClass[] */
+            public $foo;
+        };
+
+        $this->markTestSucceeded();
+
+        $this->expectException(ValueObjectException::class);
+
+        new class(['foo' => [new DummyClass(), new DummyClass(), new OtherClass()]]) extends ValueObject {
```
brendt: Again, one element should suffice.

tests/ValueObjectTest.php
```diff
@@ -125,6 +126,35 @@ public function classes_are_supported()
...
+        };
+    }
+
+    /** @test */
+    public function an_exception_is_thrown_for_a_generic_collection_of_null()
+    {
+        $this->expectException(ValueObjectException::class);
+
+        new class(['foo' => [null]]) extends ValueObject {
+            /** @var null[] */
```
brendt: This isn't what I meant with "null should not be supported", I rather meant this:

```php
new class(['foo' => [null]]) extends ValueObject {
    /** @var string[] */
    public $foo;
};
```

Meaning that typed arrays may not contain `null` values.

src/Property.php
bhulsman: That is a lot better indeed! I wasn't sure I could alter the `preg_match` usage in `resolveTypeDefinition`. After adding the `[]` it worked as expected.

tests/ValueObjectTest.php
bhulsman: Ah, was thinking to literal I guess :p

bhulsman:
Thanks for your feedback! 👍 If you have any more comments on the new changes I'd like to know.

src/Property.php
```diff
@@ -67,7 +67,7 @@ protected function resolveTypeDefinition()
             return;
         }
 
-        preg_match('/\@var ([\w|\\\\]+)/', $docComment, $matches);
+        preg_match('/\@var ([\w|\\\\\[\]]+)/', $docComment, $matches);
```
brendt: I think we'd better change this regex slightly:

```
/\@var ([\w|\\\\]+(\[\])?)/
```

This way `[]` is required to be together.

bhulsman: This will add another group (so `$varDocComment = end($matches);` won't work) and it doesn't allow for multiple types using the pipe character `|`.

I'll try and get a regex pattern together that solves this.

bhulsman:
Added better regex pattern, also see https://regex101.com/r/WhhMxu/1.

brendt:
Great! Thanks for the quick response! Really appreciate it!

brendt:
One question though: are you able to resolve the merge conflict? `ValueObjectException`

If possible, please rebase your brach on develop, this keeps the git log clean 😀

bhulsman:
> One question though: are you able to resolve the merge conflict? `ValueObjectException`
> 
> If possible, please rebase your brach on develop, this keeps the git log clean 😀

I'll try and fix the merge conflict, after I found out how to update my fork 🤓 And I assume you mean the master branch, since there is no develop branch :p.

brendt:
Great! Thanks!

I'm not able to explain rebase right now, but maybe this is a good article, if you're interested: https://medium.freecodecamp.org/git-rebase-and-the-golden-rule-explained-70715eccc372

Let's merge this!

Issue #13, opened 2018-10-23
Support non-FQCNs in phpdocs


brendt:
Right now, you need to write the whole classname in the `@var` type declaration:

```php
/** @var \Spatie\ValueObject\T */
```

Imported namespaces should also be supported.

brendt:
There's no good way to do this, unfortunately. Maybe someone comes up with a good idea, but for now we'll close it.

mfn:
My suggestion then is to not close this issue:
- people are asking for it
- it's not solved
- (just because we don't have a good solution yet, doesn't mean there won't be a fix)

Just my 2c

brendt:
You're free to submit a PR with a solution though, we'll keep that open and can discuss specifics there. I just don't think of this as "an issue": we've thought about this before, and came to the conclusion that we don't want to add support for it in the library, unless someone comes up with a better idea. That better idea belongs in a PR 🙂

Sidenote: when we discussed this a week ago, this package was still only recently released. I realise that since than, more people have started looking at it and I understand it should be an active discussion again. So let's continue this in the original PR, which I've reopened: https://github.com/spatie/data-transfer-object/pull/14

PR #14, opened 2018-10-23
Support fqcn aliasses

brendt:
Addresses #13

### DIFF
--- a/tests/ValueObject/ReflectionClassTest.php
+++ b/tests/ValueObject/ReflectionClassTest.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Spatie\ValueObject\Tests\ValueObject;
+
+use Spatie\ValueObject\ValueObject;
+use Spatie\ValueObject\ValueObjectDefinition;
+use Spatie\ValueObject\Tests\TestCase;
+use Spatie\ValueObject\Tests\TestClasses\DummyClass;
+
+class ReflectionClassTest extends TestCase
+{
+    /** @test */
+    public function it_parses_use_statements()
+    {
+        $valueObject = new class ([]) extends ValueObject {};
+
+        $class = new ValueObjectDefinition($valueObject);
+
+        $this->assertEquals(DummyClass::class, $class->resolveAlias('DummyClass'));
+    }
+}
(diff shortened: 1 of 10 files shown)


brendt:
I'm not sure we should do this. It's a lot of code for a tiny little feature, which goes against many doc block guidelines. 

Not to mention the impact this has on performance: every value object now must open and scan its class file, that's a lot of file IO..

brendt:
I don't think we want to do this at the moment. There's just way too much overhead.

brendt:
@mfn let's continue our discussion here 🙂

brendt:
There's two issue I already of with this PR:

- `use as` is not supported yet
- every DTO will re-read its source file, even if these DTOs are the same. This becomes highly inefficient in collections.

brendt:
screen shot 2018-11-08 at 10 02 20

Here's another issue with in IDEs. This import statement is regarded as unused..

mfn:
> * every DTO will re-read its source file, even if these DTOs are the same. This becomes highly inefficient in collections.

Would a solution using a private static member to cache this be deemed appropriate? …

…under the following assumption: the source never changes (as in: no one mucks with the already loaded in-memory class, i.e. no runkit-shenanigan, etc.)

brendt:
@mfn static caching is an option. Or we could have a separate cache store, two valid solutions.

tests/ValueObject/ReflectionClassTest.php
```diff
@@ -0,0 +1,22 @@
...
+{
+    /** @test */
+    public function it_parses_use_statements()
+    {
+        $dataTransferObject = new class([]) extends DataTransferObject {
+        };
+
+        $class = new DataTransferObjectDefinition($dataTransferObject);
+
+        $this->assertEquals(DummyClass::class, $class->resolveAlias('DummyClass'));
```
peter279k: It should use the `assertInstanceOf` rather than `assertEquals` assertion :).

peter279k:
The `tests/DataTransferTransferObjectTest.php` has conflicts.

It should be fixed.

brendt:
After using the package in a lot of projects, I'm now convinced there's little value in adding this feature. It's at least not worth the performance cost.

PR #21, opened 2018-10-27
Throw exception on uninitialized non-nullable property.

### LINKED ISSUE #8: Uninitialized support
Uninitialized values are now set to `null`: https://github.com/spatie/value-object/blob/master/src/ValueObject.php#L23

To be more in line with the type system that PHP 7.4 brings, we should add another state to the property, so that there's a difference between `null` and `uninitialized`.

It should be something like this:

```php
    if (
        ! isset($parameters[$property->getName()]) 
        && ! $property->isNullable
    ) {
        throw ValueObjectException::uninitialized($property);
    }

    $value = $parameters[$property->getName()] ?? null;
    
    $property->set($value);
}
```

Small remark: `ValueObjectException::uninitialisedProperty` may be renamed to `ValueObjectException::uninitialized`.

hulkur:
Resolves #8.

Changed default value for `Property::$isNullable` to `true` to be consistent with test `empty_type_declaration_allows_everything`.

Added test to allow only explicitly specified nullable property.

### DIFF
--- a/src/Property.php
+++ b/src/Property.php
@@ -19,7 +19,7 @@ class Property extends ReflectionProperty
     protected $hasTypeDeclaration = false;
 
     /** @var bool */
-    protected $isNullable = true;
+    protected $isNullable = false;
 
     /** @var bool */
     protected $isInitialised = false;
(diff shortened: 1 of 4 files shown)


src/Property.php
```diff
@@ -19,7 +19,7 @@ class Property extends ReflectionProperty
     protected $hasTypeDeclaration = false;
 
     /** @var bool */
-    protected $isNullable = false;
+    protected $isNullable = true;
```
brendt: Why did you change the default value to `true`? I prefer this package to be non-nullable by default, als it prevents lots of bugs.

hulkur: As explained in PR: to be consistent with test `empty_type_declaration_allows_everything`.
If it is left `false`, then this test breaks.

hulkur: Ok, will take a look

brendt:
Looks good! I'll merge this later today or tomorrow.

brendt:
Thanks again!

PR #23, opened 2018-10-28
Use opcode optimised methods

vanbrabantf:
## What changed?
I've updated relevant methods to their qualified namespace.

### Why?
This gives a tiny performance boost. This way PHP doesn't have to do a lookup for the namespace as it is already defined.

### This doesn't seem worth it 
It probably isn't, but as this is a package that has the potential to be used in loads of places in more significant projects these small performance boosts might add up.

You can find more info here: https://github.com/Roave/FunctionFQNReplacer

## How to test?
I've have ran the tests and did a small functional test. Everything seems to work

### DIFF
--- a/src/Arr.php
+++ b/src/Arr.php
@@ -20,7 +20,7 @@ class Arr
     {
         $keys = (array) $keys;
 
-        if (count($keys) === 0) {
+        if (\count($keys) === 0) {
             return $array;
         }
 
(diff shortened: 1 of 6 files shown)


freekmurze:
Thanks for this PR, but we’re not going to merge it in. I value simplicity of code above micro-optimizations.

Issue #33, opened 2018-11-07
toArray doens't consider nested DTOs


mfn:
Example by taking `nested_dtos_are_automatically_cast_from_arrays_to_objects` and adding a `toArray` dump:
```php
        $object = new NestedParent($data);
        var_dump($object->toArray());exit(1);
```

### Got
```
array(2) {
  ["child"]=>
  object(Spatie\DataTransferObject\Tests\TestClasses\NestedChild)#16 (4) {
    ["name"]=>
    string(5) "child"
    ["allValues":protected]=>
    array(1) {
      ["name"]=>
      string(5) "child"
    }
    ["exceptKeys":protected]=>
    array(0) {
    }
    ["onlyKeys":protected]=>
    array(0) {
    }
  }
  ["name"]=>
  string(6) "parent"
}
```
### Expected
```
array(2) {
  ["name"]=>
  string(6) "parent"
  ["child"]=>
  array(1) {
    ["name"]=>
    string(5) "child"
  }
}
```

brendt:
Are you willing to create a PR for this?

brendt:
I also added support for `toArray` on collections. These changes are tagged as 1.6.1: https://github.com/spatie/data-transfer-object/releases/tag/1.6.1

PR #36, opened 2018-11-11
Fix toArray method for nested DTO's

dzhiryakov:
Pull request for bug founded in issue #33

### DIFF
--- a/src/DataTransferObject.php
+++ b/src/DataTransferObject.php
@@ -96,7 +96,7 @@ abstract class DataTransferObject
 
     /**
      * @param mixed $property
-     * 
+     *
      * @return mixed
      */
     protected function convertForArray($property)
@@ -108,7 +108,7 @@ abstract class DataTransferObject
         if (is_object($property) && method_exists($property, 'toArray')) {
             return $property->toArray();
         }
-        
+
         if (is_array($property)) {
             foreach ($property as $key => $nested) {
                 $property[$key] = $this->convertForArray($nested);
--- a/src/DataTransferObject.php
+++ b/src/DataTransferObject.php
@@ -82,10 +82,40 @@ abstract class DataTransferObject
     public function toArray(): array
     {
         if (count($this->onlyKeys)) {
-            return Arr::only($this->all(), $this->onlyKeys);
+            $array = Arr::only($this->all(), $this->onlyKeys);
+        } else {
+            $array = Arr::except($this->all(), $this->exceptKeys);
         }
 
-        return Arr::except($this->all(), $this->exceptKeys);
+        foreach ($array as $key => $property) {
+            $array[$key] = $this->convertForArray($property);
+        }
+
+        return $array;
+    }
+
+    /**
+     * @param mixed $property
+     * 
+     * @return mixed
+     */
+    protected function convertForArray($property)
+    {
+        if ($property instanceof DataTransferObject) {
+            return $property->toArray();
+        }
+
+        if (is_object($property) && method_exists($property, 'toArray')) {
+            return $property->toArray();
+        }
+        
+        if (is_array($property)) {
+            foreach ($property as $key => $nested) {
+                $property[$key] = $this->convertForArray($nested);
+            }
+        }
+
+        return $property;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/DataTransferObject.php
```diff
@@ -82,10 +82,40 @@ public function except(string ...$keys): DataTransferObject
     public function toArray(): array
     {
         if (count($this->onlyKeys)) {
-            return Arr::only($this->all(), $this->onlyKeys);
+            $array = Arr::only($this->all(), $this->onlyKeys);
```
brendt: We can combine this if/else check into one:

```php
$array = Arr::only(
    Arr:except($this->all, $this->exceptKeys), 
    $this->onlyKeys
);
```

I wanted to refactor this for a while now, but you might as well do it in this PR.

src/DataTransferObject.php
```diff
@@ -82,10 +82,40 @@ public function except(string ...$keys): DataTransferObject
...
+
+        return $array;
+    }
+
+    /**
+     * @param mixed $property
+     *
+     * @return mixed
+     */
+    protected function convertForArray($property)
```
brendt: I'm not 100% certain about the name here. Maybe `convertPropertyToArray`? Feel free to share your opinion!

src/DataTransferObject.php
```diff
@@ -82,10 +82,40 @@ public function except(string ...$keys): DataTransferObject
...
+    {
+        if ($property instanceof DataTransferObject) {
+            return $property->toArray();
+        }
+
+        if (is_object($property) && method_exists($property, 'toArray')) {
```
brendt: Should we allow this much magic? My guess is that you have Laravel projects in mind, but I think making the assumption that everything should be cast to an array can be quite dangerous..

tests/DataTransferTransferObjectTest.php
```diff
@@ -248,4 +248,40 @@ public function nested_dtos_are_automatically_cast_from_arrays_to_objects()
         $this->assertEquals('parent', $object->name);
         $this->assertEquals('child', $object->child->name);
     }
+
+    /** @test */
+    public function nested_dtos_are_recursive_cast_from_object_to_array_when_to_array()
```
brendt: Add a separate test for every possible path, instead of combining three cases into one.

src/DataTransferObject.php
dzhiryakov: Ok

src/DataTransferObject.php
dzhiryakov: I'm agree. Naming is a pain) Your suggest is better.

src/DataTransferObject.php
dzhiryakov: I was think about this. In my personal cases it will be very useful. But for common purposes it maybe very unpredictable. It will be removed

tests/DataTransferTransferObjectTest.php
dzhiryakov: ok

src/DataTransferObject.php
dzhiryakov: Hm...it's a breaking change. By default $onlyKeys property is a []. Arr::only with empty second parameter always will be return empty array.

I suggest this option:

```php
$array = Arr::except(
     count($this->onlyKeys) ? Arr::only($this->all(), $this->onlyKeys) : $this->all(),
     $this->exceptKeys
);
```

brendt: Ah, too bad. Let's than keep it the way you implemented it for now. I find that to be clearer than the shorthand approach.

I'll probably merge this later today! 👍

brendt:
Thanks!

brendt:
Tagged as 1.6.1 (with added collection support): https://github.com/spatie/data-transfer-object/releases/tag/1.6.1

brendt:
@dzhiryakov FYI I added some more things after your PR, maybe you're interested to see the implementation? Basically DTO collections weren't supported yet, and I made some changes to the `covertToArray` method.

Here you go, if you're interested: https://github.com/spatie/data-transfer-object/commit/58cad8f31e568d709d71c6e16f008b1140bc831f

PR #37, opened 2018-11-12
Wrap checks in assert

assertchris:
See http://php.net/manual/en/ini.core.php#ini.zend.assertions and [reddit](https://www.reddit.com/r/PHP/comments/9wbcgw/strongly_typed_data_transfer_objects_to_handle/e9jzzld/?st=joejg7wx&sh=48f51c0f).

### DIFF
--- a/src/DataTransferObject.php
+++ b/src/DataTransferObject.php
@@ -21,10 +21,10 @@ abstract class DataTransferObject
     public function __construct(array $parameters)
     {
         // see http://php.net/manual/en/ini.core.php#ini.zend.assertions
-        assert($this->isValid($parameters));
+        assert($this->isValidWith($parameters));
     }
 
-    private function isValid(array $parameters): bool
+    private function isValidWith(array $parameters): bool
     {
         $class = new ReflectionClass(static::class);
 
--- a/src/DataTransferObject.php
+++ b/src/DataTransferObject.php
@@ -20,6 +20,12 @@ abstract class DataTransferObject
 
     public function __construct(array $parameters)
     {
+        // see http://php.net/manual/en/ini.core.php#ini.zend.assertions
+        assert($this->isValid($parameters));
+    }
+
+    private function isValid(array $parameters): bool
+    {
         $class = new ReflectionClass(static::class);
 
         $properties = $this->getPublicProperties($class);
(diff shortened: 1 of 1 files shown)


mfn:
Validations aren't assertions.

You may very well want validation in production and NOT activate assertions (which are global and may suddenly other parts/libraries having assertions, too).

Also: these validations (throwing an exceptions) are deliberately used for control-flow in the current implementation, which would alter in production suddenly.

TL;DR: I don't think this project is the right case for these kind of assertion wrapping.

assertchris:
Hey @mfn, thanks for your feedback.

> Validations aren't assertions.

> You may very well want validation in production and NOT activate assertions

This is fair. I think it's a balance between performance (which is the reason this approach was suggested, if you take a look at the reddit thread) and usefulness. Perhaps the default behaviour could be to validate as previously, with optional configuration to wrap it in an assertion.

brendt:
Hey Chris, thanks for the PR! I've got a few questions and thoughts.

First a technical one: will the values be set if assertions are disabled? I suppose not, as `$property->set($value);` is called within `isValidWith`?

Next a deeper question. This PR addresses the core idea of what this package is trying to do. If we see it as a mere development tool, an aid for developers to get autocompletion; than `assert` should be the way to go. Personally though, I also _want_ this package to crash on production. I want bugs, type errors in production to be reported, and not hidden away. 

Assert offers the possibility of both, you _can_ keep it enabled; though it feels like the package would be promote itself more like "the developer tool" instead of "a type system which helps prevent bugs".

What are your thoughts on this? Am I being too philosophical? Is this my opinionated "strong type side" speaking? Should I just use Rust instead of PHP for things like this? 😁

assertchris:
> First a technical one: will the values be set if assertions are disabled?

You are correct, they will not. 

> I also want this package to crash on production

Then assert is not the way to go.

> What are your thoughts on this?

Code-gen is the way to go!

mfn:
> Code-gen is the way to go!

I like that! @assertchris Can you expand on some ideas you have about this?

assertchris:
A mix between https://github.com/preprocess/pre-class-accessors and https://github.com/preprocess/pre-immutable-classes. Maybe just a light version of class-accessors, without the custom getters/setters/unsetters, and with generated docblocks and/or 7.4 property types.

brendt:
I'll close it, but everyone is free to keep the discussion going here.

PR #40, opened 2018-11-16
Using assertContains to enhance assertions

peter279k:
# Changed log
- Using `assertContains` to check the specified message contains the expected string.


brendt:
Thanks!

PR #43, opened 2018-12-03
Fix issue when requiring float in the DTO.

kevinsmith:
PHP's `gettype()` returns "double" for float, causing an exception to be thrown for valid floats with something like:

> Invalid type: expected ExampleDto::floatVariable to be of type float, instead got value `5.1`.

Example class:

```php
class ExampleDto extends DataTransferObject
{
    /** @var float */
    public $floatVariable;
}
```

This change fixes the issue by providing a mapping from "double" to "float".

### DIFF
--- a/src/Property.php
+++ b/src/Property.php
@@ -12,6 +12,7 @@ class Property extends ReflectionProperty
     protected static $typeMapping = [
         'int' => 'integer',
         'bool' => 'boolean',
+        'float' => 'double',
     ];
 
     /** @var \Spatie\DataTransferObject\DataTransferObject */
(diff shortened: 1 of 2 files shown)


mfn:
Can you please add a test, too?

brendt:
Thanks @kevinsmith ! Are you able to add a test? If not I'll add it myself.

kevinsmith:
Yup, test added!

brendt:
Thanks!

brendt:
Tagged as 1.6.6: https://github.com/spatie/data-transfer-object/releases/tag/1.6.6

PR #45, opened 2019-01-31
Adds automatic casting of DTO arrays.

### LINKED ISSUE #44: Casting of arrays
it would be nice if following was also supported:

```php
class CustomerDto
{
    /**
     * @var string
     */
    public $name;

    /**
     * @var \AddressDto[]
     */
    public $addresses;
}

$customer = new CustomerDto([
    'name' => 'John Doe',
    'addresses' => [
        ['street' => 'Street A'],
        ['street' => 'Street B'],
        ['street' => 'Street C'],
    ],
]);
```

TBPixel:
This resolves #44.

This commit adds support for automatically casting of arrays to arrays of DTOs. Example from the update README:

```php
class TagData extends DataTransferObject
{
    /** @var string */
   public $name;
}

{
    /** @var \TagData[] */
   public $tags;
}
```

`PostData` will automatically construct tags like such:

```php
$postData = new PostData([
    'tags' => [
        ['name' => 'foo'],
        ['name' => 'bar']
    ]
]);
```

### DIFF
--- a/src/Arr.php
+++ b/src/Arr.php
@@ -60,11 +60,4 @@ class Arr
 
         return array_key_exists($key, $array);
     }
-
-    public static function isAssoc(array $arr): bool
-    {
-        $stringKeys = array_filter(array_keys($arr), 'is_string');
-
-        return count($stringKeys) > 0;
-    }
 }
--- a/src/Arr.php
+++ b/src/Arr.php
@@ -60,4 +60,11 @@ class Arr
 
         return array_key_exists($key, $array);
     }
+
+    public static function isAssoc(array $arr): bool
+    {
+        $stringKeys = array_filter(array_keys($arr), 'is_string');
+
+        return count($stringKeys) > 0;
+    }
 }
--- a/src/Property.php
+++ b/src/Property.php
@@ -30,6 +30,9 @@ class Property extends ReflectionProperty
     /** @var array */
     protected $types = [];
 
+    /** @var array */
+    protected $arrayTypes = [];
+
     public static function fromReflection(DataTransferObject $valueObject, ReflectionProperty $reflectionProperty)
     {
         return new self($valueObject, $reflectionProperty);
@@ -47,7 +50,7 @@ class Property extends ReflectionProperty
     public function set($value)
     {
         if (is_array($value)) {
-            $value = Arr::isAssoc($value) ? $this->cast($value) : $this->castArray($value);
+            $value = $this->shouldBeCastToCollection($value) ? $this->castCollection($value) : $this->cast($value);
         }
 
         if (! $this->isValidType($value)) {
@@ -97,6 +100,7 @@ class Property extends ReflectionProperty
         $varDocComment = end($matches);
 
         $this->types = explode('|', $varDocComment);
+        $this->arrayTypes = str_replace('[]', '', $this->types);
 
         $this->isNullable = strpos($varDocComment, 'null') !== false;
     }
@@ -143,18 +147,11 @@ class Property extends ReflectionProperty
         return new $castTo($value);
     }
 
-    protected function castArray(array $values)
+    protected function castCollection(array $values)
     {
-        foreach ($values as $value) {
-            if (! is_array($value)) {
-                return $this->cast($values);
-            }
-        }
-
         $castTo = null;
-        $flatTypes = str_replace('[]', '', $this->types);
 
-        foreach ($flatTypes as $type) {
+        foreach ($this->arrayTypes as $type) {
             if (! is_subclass_of($type, DataTransferObject::class)) {
                 continue;
             }
@@ -168,14 +165,24 @@ class Property extends ReflectionProperty
             return $values;
         }
 
-        $casts = array_map(
-            function ($value) use ($castTo) {
-                return new $castTo($value);
-            },
-            $values
-        );
+        $casts = [];
 
-        return ! empty($casts) ? $casts : null;
+        foreach ($values as $value) {
+            $casts[] = new $castTo($value);
+        }
+
+        return $casts;
+    }
+
+    protected function shouldBeCastToCollection(array $values): bool
+    {
+        foreach ($values as $value) {
+            if (! is_array($value)) {
+                return false;
+            }
+        }
+
+        return true;
     }
 
     protected function assertTypeEquals(string $type, $value): bool
--- a/src/Property.php
+++ b/src/Property.php
@@ -47,7 +47,7 @@ class Property extends ReflectionProperty
     public function set($value)
     {
         if (is_array($value)) {
-            $value = $this->cast($value);
+            $value = Arr::isAssoc($value) ? $this->cast($value) : $this->castArray($value);
         }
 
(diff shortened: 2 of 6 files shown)


src/Property.php
```diff
@@ -143,6 +143,39 @@ protected function cast($value)
...
+        if (! $castTo) {
+            return $values;
+        }
+
+        return array_map(
+            function ($value) use ($castTo) {
+                if (is_array($value)) {
+                    return new $castTo($value);
+                }
+            },
```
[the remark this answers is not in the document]
TBPixel: Ah, that's a good gotcha! To walk through my thinking: I wanted to avoid attempting to pass a non-array to the constructor of a DTO as that's the only type it it accepts. However, I agree that I don't think this is the correct approach in that it casts too wide of a net.

Perhaps a better approach would be to first validate that _all_ of the values inside of a value array were just arrays, and if one isn't throw a `DataTransferObject::invalidType` exception.

src/Property.php
```diff
@@ -143,6 +143,41 @@ protected function cast($value)
         return new $castTo($value);
     }
 
+    protected function castArray(array $values)
+    {
+        foreach ($values as $value) {
```
brendt: Can you explain why this loop is here? It feels like it should be done on a higher level, maybe together with the `Arr::isAssoc($value)` check?

src/Property.php
```diff
@@ -143,6 +143,41 @@ protected function cast($value)
         return new $castTo($value);
     }
 
+    protected function castArray(array $values)
```
brendt: What about renaming this method to `castCollection` to differentiate between "a collection of DTOs" and a simple array?

src/Property.php
brendt: Thinking about it, perhaps we shouldn't use `Arr::isAssoc($value)` but instead a separate function called something like "shouldBeCastToCollection", which would loop once over all items, and check whether

- The keys are no strings, to replace `isAssoc`
- The values are arrays

src/Property.php
```diff
@@ -143,6 +143,41 @@ protected function cast($value)
...
+    {
+        foreach ($values as $value) {
+            if (! is_array($value)) {
+                return $this->cast($values);
+            }
+        }
+
+        $castTo = null;
+        $flatTypes = str_replace('[]', '', $this->types);
```
brendt: Perhaps it'd be better to parse the flatTypes in `Property:: resolveTypeDefinition`, so that we don't have to do all this "setup work" while casting.

src/Property.php
```diff
@@ -143,6 +143,41 @@ protected function cast($value)
...
+            $castTo = $type;
+
+            break;
+        }
+
+        if (! $castTo) {
+            return $values;
+        }
+
+        $casts = array_map(
```
brendt: Again I think a loop would be the better choice here, in light of performance.

src/Property.php
```diff
@@ -143,6 +143,41 @@ protected function cast($value)
...
+            function ($value) use ($castTo) {
+                return new $castTo($value);
+            },
+            $values
+        );
+
+        return ! empty($casts) ? $casts : null;
```
brendt: Empty casts means there were no items passed, right?

```php
$postData = new PostData([
    'tags' => []
]);
```

Or maybe even

```php
$postData = new PostData([]);
```

In the first case, I believe an empty array should be returned, not `null`. The second case should only be allowed if `tags` is nullable, but this might already work as is, you could test that.

src/Property.php
mfn: I did some microbenchmarks using this in `\Spatie\DataTransferObject\Property::set`:
```php
            $start = microtime(true);
            for ($i = 0; $i < 1000000; $i++) {
                $this->castArray($value);
            }
            var_dump(microtime(true) - $start);exit(1);
```
I've then switch between this PR and a simply foreach loop:
```php
        $tmp = [];
        foreach ($values as $value) {
            $tmp[] = is_array($values)
                ? new $castTo($value)
                : null;
        }
        return $tmp;
```

With `foreach` (after a few runs, i.e. "hot system"): consistently `3.2s`

With `array_map`: `3.4s`

(Note: this was benchmark before the latest force-push which changed the method slightly)

src/Property.php
TBPixel: So, initially the goal of that loop was to validate that the array is an array of arrays. I agree with your point though that this validation should be separated out. Renaming it, especially in context with your further comments, also helps contribute to the consistency of the naming within the package; I'll make sure to include that in my next update.

src/Property.php
TBPixel: No problem, I think that increases the consistency of the naming convention in the package.

src/Property.php
TBPixel: While performance is clearly a non issue, I'm inclined to feel that a loop is easier to read, follow, and make changes to in future; I believe it's the better approach in this instance as opposed to `array_map`.

src/Property.php
TBPixel: I believe the second case, as you said, does already work however I am unsure on the first. I should add additional test(s) and correct this if it's inconsistent with expected behavior.

TBPixel:
Hey, thanks for all the feedback! I'm new to making open source contributions so I really appreciate it the effort in this code review. I'll try to incorporate and learn from the changes being suggested!

src/Property.php
TBPixel: This seems like a better place to put the responsibility of the type definition, yes.

src/Arr.php
```diff
@@ -60,4 +60,11 @@ public static function exists($array, $key): bool
 
         return array_key_exists($key, $array);
     }
+
+    public static function isAssoc(array $arr): bool
+    {
+        $stringKeys = array_filter(array_keys($arr), 'is_string');
```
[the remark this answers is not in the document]
TBPixel: This is good feedback to get me thinking, but it seems in a later comment the better suggestion of removing this entirely in favour of a more specific `shouldBeCastToCollection` method takes priority.

TBPixel:
I've made a number of changes to address these issues.

- I removed the `Arr::isAssoc` method in favour of a `Property->shouldCastToArray` method.
- I added a new test case to validate nullable casted collections.
- I added a `Property::$arrayTypes` attribute. This is resolved in the `Property->resolveTypeDefinition` method and is separated from the existing `$types` attribute because I wanted to be sure there was no conflict (eg. a DTO collection accidentally allowing the base type as a valid value).
- I renamed the `castArray` method to `castCollection`, removed array_map in favour of a loop.

TBPixel:
I missed re-implementing the associative array check in the new `shouldCastToCollection` method, which naturally caused a type error bug during casting. I also found that the null dto test was somewhat invalidated by setting a default value on my initial `NestedParentOfMany` classes children (equal to an array) so I fixed that.

kminek:
@TBPixel great work mate! hope this is merged soon!

brendt:
Great work, Tony! Let's get this merged.

brendt:
And finally, tagged as 1.7.0: https://github.com/spatie/data-transfer-object/releases/tag/1.7.0

PR #48, opened 2019-02-11
Fixes #47 - Allowing empty DTO's to be cast via empty arrays

### LINKED ISSUE #47: Invalid type exception when casting empty array
Following code shoud not throw an exception:

```php
class ParentDto extends DataTransferObject
{
    /**
     * @var \ChildDto
     */
    public $child;
}

class ChildDto extends DataTransferObject
{
}

$child = new ChildDto([]);

$parent = new ParentDto([
    'child' => $child,
]);

dump($parent->toArray());

$parentFromArray = new ParentDto([
    'child' => [],
]);
```

TBPixel:
This adds a new `EmptyChild` DTO test class and associated test to resolve #47. Empty DTO's can be cast to using an empty array now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to `data-transfer-object` will be documented in this file
 
+## 1.7.1 - 2019-02-11
+
+- Fixes #47, allowing empty dto's to be cast to using an empty array.
+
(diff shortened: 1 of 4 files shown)


brendt:
Thanks for the quick response, really appreciate it!

PR #50, opened 2019-03-15
Immutable data transfer objects

### LINKED ISSUE #12: Readonly support
I'm not sure how we'd go about this one. But properties on value objects should be read only. I don't think `__set` is of any help, as this are public properties..

brendt:
By using `ImmutableDataTransferObject` as a proxy, we're able to have the best of both worlds!

closes #12

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to `data-transfer-object` will be documented in this file
 
+## 1.8.0 - 2019-03-??
+
+- Support immutability
+
 ## 1.7.1 - 2019-02-11
 
 - Fixes #47, allowing empty dto's to be cast to using an empty array.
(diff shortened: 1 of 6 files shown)


brendt:
One obvious shortcoming I can already think of: you can still set values on nested values. Eg:

```php
$immutable->childObject->property = 'abc';
```

We could chain the immutable calls to all nested DTOs, but this still doesn't solve the problem for other kind of objects and array.

sebastiandedeyne:
> One obvious shortcoming I can already think of: you can still set values on nested values.

On the other hand, this is the default behaviour of `clone` in PHP so I don't think it's a huge issue.

brendt:
Freek opted the idea to make this the default behaviour in the next major version: https://github.com/spatie/data-transfer-object/issues/52

Issue #52, opened 2019-03-18
Immutable by default


brendt:
We could make immutable DTOs the default in the next major version. Mutabel DTOs would be possible with the `::mutable` static constructor.

deviouspk:
I agree Dto's should be immutable by default. I don't even see a use case (other than bad practice) where you would want it to be mutable.

deviouspk:
I just checked the immutable implementation and i think you're right. There's probably no way to do this without losing the type hint functionality. Until https://wiki.php.net/rfc/immutability hits i don't think you can create immutable attributes without proxying them through to another class :(

deviouspk:
Ok so after coding the feature for validation i found a way to do the "impossible" and make it immutable without losing the typehinting. We can store the properties in an array. and unset all the public variables: 

```
        $properties = $this->getPublicProperties($class);
        foreach ($properties as $property) {
            $this->properties[] = $property;
            unset($this->{$property->getName()});
        }
```

Then we add that magic __set method to the default datatransferobject class that throws an error (to block setting) & a magic __get method to proxy through the getting to the $properties array. And boom we have an immutable dto with ide typehinting :)

deviouspk:
#55 proof of concept

brendt:
Ah you wrote the explanation here, I must say I'm impressed. Could you share the explanation also in the PR? This way we can continue the conversation over there.

deviouspk:
Hey brendt i see this issue is still open. Could you please checkout the implementation on https://github.com/larapie/data-transfer-object. 

Should give you some good clues to implement immutability without compromises.
I'm gonna maintain this package with key differences like validation, annotations, ..

brendt:
I still plan on adding immutability in the future, no ETA yet. As said in previous PRs and issues: I'm open for a PR that only addresses the immutability part and follows our package guidelines.

PR #55, opened 2019-03-25
Immutable by default + start validation

deviouspk:
Do not merge. This pull request is purely to show the concept for immutable dtos without loss of typehinting  #52 & a new feature data validation !

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ composer.lock
 docs
 vendor
 coverage
+.idea
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
     },
     "require-dev": {
         "larapack/dd": "^1.0",
-        "phpunit/phpunit": "^8.0"
+        "phpunit/phpunit": "^7.0"
     },
     "autoload": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
     },
     "require-dev": {
         "larapack/dd": "^1.0",
-        "phpunit/phpunit": "^7.0"
+        "phpunit/phpunit": "^8.0"
     },
     "autoload": {
         "psr-4": {
--- a/src/Attribute.php
+++ b/src/Attribute.php
@@ -1,0 +1,84 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: arthur
+ * Date: 25.03.19
+ * Time: 20:15
+ */
+
+namespace Spatie\DataTransferObject;
+
+
+use Closure;
+
+class Attribute
+{
+    /**
+     * @var Attribute
+     */
+    protected $property;
+
+    /**
+     * PropertyFactory constructor.
+     * @param Property $property
+     */
+    public function __construct(Property $property)
+    {
+        $this->property = $property;
+    }
+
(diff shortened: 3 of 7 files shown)


brendt:
A quick tip: you can make draft PRs in GitHub now :)

Can you explain how this PR solves the issue of immutability?

deviouspk:
To achieve this "immutable dto" without the loss of ide typehinting we basically remove the public properties from the class on runtime and add the properties to an array that we access through the magic __get method. Followed by an error thrown in the magic __set method when we try to set a variable in the dto.

```
        $properties = $this->getPublicProperties($class);
            $this->properties[] = $property;
            unset($this->{$property->getName()});
        }
```

```
    public function __get($name)
    {
        return $this->properties[$name]->getActualValue();
    }
```
```
    public function __set($name, $value)
    {
        throw DataTransferObjectError::immutable($name);
    }

```

The validation feature is implemented through a factory approach for the property class. This allows us to validate the properties inside of the dto. Since type validation is already done inside the dto i honestly think this doesn't violate the single responsability principle any more than the validation of the type already does.

This is how a dto with validation could look like:

class ValidateableDto extends DataTransferObject
{
    /** @var string */
    public $name;

    /** @var int */
    public $number;
    
    public function name(Attribute $attribute){
        return $attribute
            ->optional()
            ->default("arthur")
            ->rule("min:2")
            ->rule("max:30");
    }
    
    public function number(Attribute $attribute){
        return $attribute
            ->nullable()
            ->constraint(function ($value){
                return $value>5;
            });
    }
}

I would love to get some input on both feature implementations. Ofcourse this pull request is a quick proof of concept. If the features get accepted i'll gladly do it properly.

Best regards,
Arthur

deviouspk:
@brendt Please take a look at the code now. Cleaned it a little.

deviouspk:
For the validation would you be ok with including the laravel validation package?

https://github.com/illuminate/validation

brendt:
I still find the validation aspect a difficult consideration. As you rightly suggest: it would be insane to write a validation layer from scratch in this package. On the other hand: it doesn't seem the best idea to include Laravel validation. This is a framework agnostic package, and I believe we should keep it that way.

Having thought about it some more, I think the best approach would be to make a framework specific package (Laravel in your case?) which wraps this DTO package. Since we don't need this feature in our projects at Spatie, we're not going to invest time in it.

My suggestion would be that if you made a framework specific package, we could mention it in the readme as a suggestion.

I'm still very interested in the immutable part, but that should be a separate PR, are you ok with making that one? Could you also make sure you keep the same coding style? We don't add redundant docblocks like these: https://github.com/spatie/data-transfer-object/pull/55/files#diff-8ce7c5bc0851d7f504f56b76c7af99e2R195

deviouspk:
Hey brendt. 
Please checkout https://github.com/larapie/data-transfer-object if you have the time.

I'm gonna maintain that repo and try to keep it in sync if you introduce new features. 
It introduces a bunch of new features through annotations supported by doctrine/annotations:
- Validation (through annotations with symfony/validation )
- Optional properties (through annotations)
- Overriding/adding Properties on runtime.
- Immutable Properties (through annotations)

PR #56, opened 2019-03-27
Immutable by default + support for mutation of methods for external packages

deviouspk:
This immutable dto without the loss of ide typehinting is achieved through removing the public properties from the value object on runtime and adding the properties to an array that we access through the magic __get method. 

```
        $properties = $this->getPublicProperties($class);
            $this->properties[] = $property;
            unset($this->{$property->getName()});
        }
```

```
    public function __get($name)
    {
        return $this->properties[$name]->getActualValue();
    }
```
```
    public function __set($name, $value)
    {
        throw DataTransferObjectError::immutable($name);
    }

```


brendt:
How difficult would it be to mark individual properties as immutable, on top of the general `::immutable` method?

deviouspk:
Hmm not that hard. I would do it like this:
introduce a new empty interface called immutableProperty. Then typehint for it with the docs.

```
class NestedChild extends DataTransferObject
{
    /** @var string|ImmutableProperty */
    public $name;
}
```
Then i would introduce a variable "immutable" on the property class. that gets set when the property contains that doc. And then check for it on the dto when you try to set a property on a mutable dto object. I can add it in this pull request if you want. Shouldn't take more than 5 min

deviouspk:
Oh you mean you want to declare the properties in the static mutable method?

brendt:
I was thinking something like this

```php
class NestedChild extends DataTransferObject
{
    /** @var string immutable */
    public $name;
}
```

I know we'd be "abusing" the doc blocks, but we have to keep PHP 7.4 in mind:

```php
class NestedChild extends DataTransferObject
{
    /** @var immutable */
    public string $name;
}
```

This should also work. I'm not sure if we should add it or not.. I'd like to hear your input

deviouspk:
Hmm i wouldn't break the php convention. What about this?
```
class NestedChild extends DataTransferObject
{
    /** @var string|Immutable */
    public $name;
}
```

brendt:
Yeah, let's do it your way!

```
/** @var string|immutable */
``` 

Seems good!

deviouspk:
Added the immutable property feature

deviouspk:
Wouldnt it be better if the immutability/mutability of the dto is declared on the class itself and not on runtime through a static method? We could for example make it mutable by default and if you want immutable behavior you would need to implement the immutable interface on the dto class.

src/Property.php
```diff
@@ -122,6 +136,12 @@ protected function resolveTypeDefinition()
         $this->arrayTypes = str_replace('[]', '', $this->types);
 
         $this->setNullable(strpos($varDocComment, 'null') !== false);
+
+        if (in_array('immutable', $this->types) || in_array('Immutable', $this->types)) {
```
brendt: This check for immutability must be set before 

```php
$this->hasTypeDeclaration = true;
```

The reason is PHP 7.4:

We want to allow the following:

```php
class MyDTO
{
    /** @var immutable */
    public string $foo;
}
```

In this case we want PHP's type system to handle type checking, so `$this->hasTypeDeclaration` has to be `false`. But we do want to keep the immutability support.

I think we want something like this:

```php
if (in_array('immutable', $this->types) || in_array('Immutable', $this->types)) {
    $this->isImmutable = true;

    unset(
        $this->types['immutable'],
        $this->types['Immutable']
    );
}

if (empty($this->types)) {
    return;
}

$this->hasTypeDeclaration = true;

// …
```

Also note that you can use one unset call to unset multiple variables.

src/Property.php
```diff
@@ -13,6 +13,7 @@ class Property
         'int' => 'integer',
         'bool' => 'boolean',
         'float' => 'double',
+        'immutable' => immutable::class
```
brendt: Can you explain why this interface is needed?

src/Property.php
```diff
@@ -96,6 +100,16 @@ public function setNullable(bool $bool): void
         $this->isNullable = $bool;
     }
 
+    public function isImmutable(): bool
+    {
+        return $this->isImmutable;
+    }
+
+    public function setIsImmutable(bool $isImmutable): void
```
brendt: Why do we need a setter? The variable should be set from inside the `resolveTypeDefinition` call, so we can directly call it.

src/TestClass.php
```diff
@@ -0,0 +1,15 @@
+<?php
+/**
```
brendt: Could you remove these docblocks?

src/TestClass.php
```diff
@@ -0,0 +1,15 @@
...
+ * Created by PhpStorm.
+ * User: tony
+ * Date: 27.03.19
+ * Time: 14:49
+ */
+
+namespace Spatie\DataTransferObject;
+
+
+class TestClass implements immutable
```
brendt: Can you explain exactly why this is needed?

src/immutable.php
```diff
@@ -0,0 +1,15 @@
+<?php
```
brendt: Also remove these docblocks.

src/DataTransferObject.php
```diff
@@ -24,12 +24,12 @@ abstract class DataTransferObject implements DtoContract
     /** @var bool */
     protected $immutable;
 
-    public static function mutable(array $parameters): DtoContract
+    public static function mutable(array $parameters): self
```
brendt: I'm unsure whether the `mutable` static constructor should be added. I find this confiusing:

```php
$dto = MyDTO::mutable(/* … */);

// Won't work, even though we made a mutable object
$dto->foo = 'bar'
```

_If_ we'd keep this construct, I'd expect it to make the whole object mutable. But I don't think we want to allow that.

src/DataTransferObject.php
```diff
@@ -7,92 +7,206 @@
...
+    {
+        return new static($parameters, false);
+    }
+
+    public static function immutable(array $parameters): self
+    {
+        return new static($parameters, true);
+    }
+
+    public function __construct(array $parameters, bool $immutable = true)
```
brendt: We shouldn't handle immutability via the constructor. Because this package is meant to be extended by concrete DTO classes, we want to constructor to be as clean as possible.

Let's add a fluent setter and use it like so:

```php
public static function immutable(array $parameters): self
{
    return (new static($parameters, true))->setImmutable(true);
}
```

src/DataTransferObject.php
```diff
@@ -7,92 +7,206 @@
...
+
+    /** @var bool */
+    protected $immutable;
+
+    public static function mutable(array $parameters): self
+    {
+        return new static($parameters, false);
+    }
+
+    public static function immutable(array $parameters): self
```
brendt: Returning `self` is a little problematic in PHP, because sub classes cannot return it. Here's an example:

```php
class A
{
    public function foo(): self {}
}

class B extends A
{
    public function foo(): self {}
}
```

This gives an error because the return type of `B::foo` should be `A`, instead it's `B`. That's why I always return the concrete class, to make it very clear what should be returned.

```php
class A
{
    public function foo(): A {}
}

class B extends A
{
    public function foo(): A {}
}
```

As soon as PHP 7.4 drops with better type variance, there are more possibilities, but for now this will have to do.

deviouspk:
Ok i'll go through the changes. Regarding the immutable/mutable behavior what do you think about this:

```
    /**
     * @return static
     */
    public function mutable()
    {
        $this->immutable = false;
        return $this;
    }

    /**
     * @return static
     */
    public function immutable()
    {
        $this->immutable = true;
        return $this;
    }
```

And this feature:
```
    protected function resolveImmutable()
    {
        if ($this instanceof immutable)
            $this->immutable = true;
    }
```

That would allow us to make a dto immutable on the class itself.

src/Property.php
deviouspk: It isn't was something i forgot to remove.

src/Property.php
deviouspk: yeah thanks i forgot

src/Property.php
deviouspk: for external packages who want control of things in the property class :)

src/TestClass.php
deviouspk: not i removed it

src/DataTransferObject.php
deviouspk: "As soon as PHP 7.4 drops with better type variance," I've been waiting for that for years (that along with typed properties & generics).

What about this implementation:
```
    /**
     * @return static
     */
    public function mutable()
    {
        $this->immutable = false;
        return $this;
    }

    /**
     * @return static
     */
    public function immutable()
    {
        $this->immutable = true;
        return $this;
    }
```

deviouspk: it will break the current api (that contains the immutable static method) though and set mutable by default

deviouspk:
I've cleaned up the structure of the project. Introducted an interface for property. Removed the array class and improved the implementation of the except feature (slight performance increase). Also Created separate exception classes (for improved testability). 
This should be the final push. I consider this mergeable :+1:

src/DataTransferObject.php
deviouspk: removed mutable method

deviouspk:
Added immutable support for nested dtos

src/DataTransferObject.php
deviouspk: i readded the static immutable method and marked it as deprecated

src/DataTransferObject.php
```diff
@@ -66,12 +64,37 @@ protected function boot(array $parameters): void
         }
 
         $this->processRemainingProperties($parameters);
+        $this->determineImmutability();
     }
 
-    protected function resolveImmutable()
+    protected function determineImmutability()
     {
         if ($this instanceof immutable) {
```
brendt: It's still unclear to me why we need this interface, can you explain?

src/DataTransferObjectCollection.php
```diff
@@ -7,6 +7,7 @@
 use Iterator;
 use Countable;
 use ArrayAccess;
+use Spatie\DataTransferObject\Contracts\DtoContract;
```
brendt: This extra contract seems like an overcomplication. Can you explain the reasoning behind it?

deviouspk: Until php has better type variance support this seemed like a good idea to sprcify the return type of a dto. I just added it everywhere for consistency

src/DataTransferObject.php
deviouspk: If a user wants to make a dto immutable on runtime he can do it by implementing the immutable interface. This is a nice addition for people who want immutable by default.

deviouspk:
```
class MyDto extends DataTransferObject implements immutable
{
    /** @var string */
    public $name;
}
```

brendt: Ok, I'd rename `immutable` to `Immutable` though

deviouspk: Sure.

brendt:
@deviouspk I appreciate the effort you're putting into this, but you're changing way too much stuff for this feature. A PR should address one feature only. I expect this PR to only add immutability support and nothing else.

Having thought about it some more, I think we should allow two things:

- First, an immutability flag per property:

```php
class DTO
{
    /** @var string|immutable */
    $foo;
}
```

- Second, an `immutable` static constructor:

```php
$dto = DTO::immutable([ /* props */]);
```

That means no extra interfaces, no other refactors, no config changes. Like I said I really appreciate the effort you're putting into this, but we simply cannot merge a PR of 700+ additions, which is basically a rewrite of half the package.

I'm really don't want to discourage you, but I'll have to close this PR. Again: if you're up for it you can submit it new one with only the changes we need to properly support immutability.

deviouspk:
No offense but this pull request only added immutability support and no other features. The only other thing i did was some refactoring wich was necessary anyway to introduce the  properties variable on the dto to support immutability and i cleaned up a lot of mess on the way. The api is backwards compatible with the old one so my pull request doesn’t break anything. Sorry but I’m not going to make a new one for the sake of “too many lines changed” It seems to me that you did not take the effort to go through the code and are scared that something will break. Anyway this pull request is ready to be merged. It’s up to you to do what you want with it. The features that you requested are in it. Including the static immutable method. Immutability without typehint loss and immutable properties. Please note that If you want something specific to change for a reason i’m willing to look at it. (For example i could remove the interfaces if thats what you want) but the downside is that it becomes less strongly typed for other third party packages that extend from this. 

I’ll be waiting for your reply.

brendt:
Again I'm sorry if I offended you, but the PR is just way too large to be able to merge. Claiming that you're cleaning up a lot of mess on the way is exactly what I asked not to do. You're always free to submit PRs to improve the code, but these changes do not belong in the same immutability PR.

I implemented in a separate branch, you can take a look if you want. It's about 200 lines of changes, instead of the 1000 of this PR: https://github.com/spatie/data-transfer-object/pull/57/files

deviouspk:
#57 is in many ways inferior, slower and less maintainable than this implementation. Can you elaborate on what part of cleaning up you didn't like?

I'll give a list of the the things that were "cleaned up":
- Decouple exceptions to seperate classes (good practice & improved testability)
- added interfaces (convenience for external packages (can be removed when better type variance is implemented in php 7.4))
- Remove except property on the dto and reimplement the feature by adding visibility to the property class (improved performance, testability + ability to remove external Arr class)
- Move logic from constructor to the boot method + refactor code there (good practice, more maintainable, overridable for third party packages)
- Create own property object instead of extending it from the php reflection api. (improved flexibility & maintainability)

brendt:
@deviouspk like I said before, it's not that I don't like your changes, they just don't belong in this PR. Try to put yourself in my place: as the maintainer I must ensure the package stays maintainable in the long run. I definitely welcome contributions, but they should be added one at a time. Reviewing a PR with 1k lines of code is very difficult to do right.

So I'd definitely would like your refactors to be added, but one at a time, not like this. 

PS: in testing my immutability PR, I've noticed that there are some breaking changes on mutable objects. This means we'll have to tag a major release in order to get this through, so if you want to propose improvements that might break backwards compatibility: this is the time :)

deviouspk:
Hmm what breaking changes are we talking about. I don't had any breaking changes. The only breaking change could be immutable by default. But i solved that by adding an immutable interface that allowed the users to choose if they wanted immutable by default or not.
Regarding the big pull request couldn't you just merge my pull request and add it as a alpha rc. I'm 99% sure everything works as expected since i use it myself already in my projects and all tests pass. But an alpha release just in case for that 1% ?

brendt:
Take a look at the following:

```php
        $dto = new class([]) extends DataTransferObject {
            /** @var array */
            public $array = [];
        };

        $dto->array[] = 'abc';
```

This results in an error

```php
Indirect modification of overloaded property class@anonymous::$array has no effect
```

deviouspk:
Hmm yes you could work around that:

```
    public function &__get($name)
    {
        return $this->properties[$name]->value;
    }
```

But this would mean that the values from arrays are not really immutable. I don't think there's a way around that

deviouspk:
hmm what you could do is make an "Immutable" Trait that overrides the __get && __set methods. If you try to alter an immutable object it will throw an immutable exception or an indirect modification error (depending on if you are for example trying to add values to an array or actually overriding the variable).
I don't think you will be able to catch the indirect modification error because:
1. it's not an exception but a warning i think.
2. it happens outside of the dto class. So there's no place to try catch anyway.

Could be wrong though didn't look into it in depth.

This implementation also implicates that you cannot make the dto immutable on runtime. 
doing $dto->setImmutable() won't work since you cannot declare traits on runtime.

It's a compromise but one that i would make. This would give the users 2 possibilities to make a dto immutable. Either on the dto class itself or through the static method immutable().

the __get method for mutable dto's would look like this:
```
    public function &__get($name)
    {
        if ($this->properties[$name]->isImmutable()) {
            return $this->properties[$name]->getValue();
        }
        return $this->properties[$name]->value;
    }
```
the one for the immutable dto:
```
    public function __get($name)
    {
        return $this->properties[$name]->getValue();
    }
```

You would also need to adjust the setter and use php reflection to check if the class actually uses the immutable trait. And modify my chain immutable code to make sure the casted dto's are also immutable.
I'm not sure if you have that immutable chaining features in your pull request.

deviouspk:
After giving this some thought i think it’s best if we follow the behavior other programming languages have aswell and this is basically not being able to make an object mutable/immutable on runtime. But ONLY on instantiation. Do you still want to make objects immutable by default? I think that would be the best imho. That would mean we deprecate the static immutable method and add a static mutable method. And create 2 traits (immutable trait and mutable trait) in wich you implement the different __get methods and in both you add a public method isImmutable. That will be used to check on runtime if we need to cast the nested dto’s to an immutable or mutable dto. Sorry if this sounds complicated :p

brendt:
I'm going to give this some more thought before settling on a solution

PR #57, opened 2019-04-01
Immutability

brendt:
Support immutability


src/Property.php
```diff
@@ -95,14 +115,25 @@ protected function resolveTypeDefinition()
...
         $varDocComment = end($matches);
 
-        $this->types = explode('|', $varDocComment);
-        $this->arrayTypes = str_replace('[]', '', $this->types);
+        $types = explode('|', $varDocComment);
+
+        $immutableIndex = array_search('immutable', $types);
+
+        if ($immutableIndex !== false) {
+            $this->isImmutable = true;
```
brendt: Use `markImmutable` here

PR #58, opened 2019-04-10
moved object hydration to separate method, added strict flag to check…

piotrgrabowskielastique:
moved object hydration to separate method, added strict flag to check for unknown properties.

### DIFF
--- a/src/DataTransferObject.php
+++ b/src/DataTransferObject.php
@@ -15,6 +15,8 @@ abstract class DataTransferObject
     /** @var array */
     protected $onlyKeys = [];
 
+    protected $strict = true;
+
     /**
      * @param array $parameters
      *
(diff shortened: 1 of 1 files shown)


brendt:
Can you explain what advantage this hydrate method has? 

Also: this package is built from the ground up on strict typing, we'll not be changing this behaviour. In its current form I won't be able to merge this PR, sorry.

piotrgrabowskielastique:
Hydration in separate method enables you to add custom behavior for non immutable DTO's. Strict flag is useful for situations where you want to build a object from data that has extra fields that you dont need (like building it from 3rd party apis).

brendt:
If you're working with third party APIs, you should manually cast the properties to the correct type. This is an intentional decision and we'll not change it. A strong type system can prevent certain bugs from happening, and that's exactly what this package is about.

If you're interested in representing API data as an object in a non strict way, you could write a small factory to convert your array of data to an object with public properties. That's outside the scope of this package though.

PR #62, opened 2019-09-13
Allow ignoring unknown parameters on DTO instantiation

### LINKED ISSUE #61: Ignoring keys that are not defined in the DTO
I'm using `spatie/data-transfer-object` to wrap the result of a third-party API.

I totally accept the risk that they might remove a key from the result payload, in which case it would break on the instantiation of my DTO, with a really clear exception.

That behavior is fine, and what I expect.

The problem is with keys that aren't defined in the DTO.

For instance, if I define the following DTO:

```php
{
    /** @var int */
    public $a;

    /** @var int */
    public $b;
}
```

And I instantiate one like so:
```php
$data = new MyDTO(['a' => 1, 'b' => 2, 'c' => 3]);
```

It crashes, indicating that I haven't explicitly declared the `c` key.

This means that if I want to use DTOs, I have to either:
- Declare all possible keys (which is annoying because that 3rd party API returns a good hundred of keys)
- Remove these keys before instantiating my DTO (which means I have to maintain two separate list of fields: one in the DTO, one for my whitelist/blacklist).

I think a good solution would be to add the possibility for  `DataTransferObject` to simply ignore the keys it isn't aware of.

----

We would simply need to condition this line https://github.com/spatie/data-transfer-object/blob/master/src/DataTransferObject.php#L51 behind some kind of flag.

How would you feel about adding such a flag:
```diff
-public function __construct(array $parameters)
+public function __construct(array $parameters, bool $ignoreUnkownParameters = false)
```

Alternatively, we could add a `ignoreUnknownParameters` to the base class, and override it in our implementations of the DTO.

rgehan:
This PR fixes #61.

It allows to declare a DTO that will explicitly ignore unknown parameters.

Such a DTO would be declared that way:

```php
{
    protected $ignoreUnknownParameters = true;

    /** @var int */
    public $a;

    /** @var int */
    public $b;
}
```

And would allow instantiating it with any array, provided it contains both `a` and `b`

```php
// Would not throw
new MyDTO(['a' => 1, 'b' => 2, 'c' => 3]);
```

Feel free to suggest any improvement.


brendt:
Please see my answer in #61

PR #63, opened 2019-09-17
Changed casted to cast in readme

koomai:
From https://writingexplained.org/cast-vs-casted:

> Casted is an incorrect past tense conjugation of the verb cast. It almost never appears in print or edited works of any kind.

Example of past tense from https://laravel.com/docs/6.x/eloquent-mutators#attribute-casting:

> Now the `is_admin` attribute will always be **cast** to a boolean when you access it...


brendt:
Ah yes, thank you!

PR #65, opened 2019-09-25
make parseArray() more obvious

Gummibeer:


mfn:
👎 on the change.

With the old code it's more clear what the default behaviour of the loop is expected to do. Also early returns/continues are _usually_ nicer to mentally follow (brain can easier short-circuit).

Gummibeer:
But the most common case is do nothing. Handling an array value is the exceptional case with special treatment.

mfn:
Fair point!

Still I consider the current version more readable 🤷‍♀️

Gummibeer:
The reason why I think that this change makes it more readable:
* no inversion - inversions have two disadvantages
  * They are harder to read because a single `!` changes the whole condition
  * They aren't this precise - `is_array()` **only** allows an array, `!is_array()` allows **everything** except an array.
* a new case can be added easier

PR #66, opened 2019-10-01
cache DTO properties

### LINKED ISSUE #59: Improve performance by memoizing property types
Hello, thankyou for this excellent library, it has saved me from writing so much boilerplate! :smile:

I'm wondering if there's any reason the property types of the DTO can't be memoized? Currently it re-parses all the doc comments every time you construct an instance, which seems like a lot of unnecessary work when you are creating thousands of instances.

I tried replacing [this line](https://github.com/spatie/data-transfer-object/blob/06704275d0be83f3891342c5b165aaeacb94e8de/src/DataTransferObject.php#L32) with:

```php
if (!isset(self::$properties)) {
    self::$properties = $this->getPublicProperties($class);
}
$properties = self::$properties;
```

And with that change, creating 10000 instances of a DTO goes from taking ~1.36 seconds to ~0.58 seconds. Here's my test script:

```php
class User extends DataTransferObject
{
    /** @var int */
    public $id;
    /** @var string */
    public $name1;
    /** @var string|null */
    public $name2;
    /** @var string|null */
    public $name3;
    /** @var string|null */
    public $name4;
    /** @var string|null */
    public $name5;
    /** @var string|null */
    public $name6;
    /** @var string|null */
    public $name7;
    /** @var string|null */
    public $name8;
    /** @var string|null */
    public $name9;
}

for ($i = 0; $i < 10000; $i++) {
    $user = new User(['id' => 123, 'name1' => 'Bob']);
}
```

Would this be a safe change to make, or is there a case where it would be a change in behaviour for some people? If it's a desirable change, we'd be happy to work on a proper PR for it. It would certainly make a difference for us, as we (a) are using it all over a large codebase and (b) map thousands of objects to DTOs in some places.

Gummibeer:
fixes #59 

The approach/idea is copied from [spatie/enum](https://github.com/spatie/enum) package.

### DIFF
--- a/src/Property.php
+++ b/src/Property.php
@@ -6,6 +6,9 @@ namespace Spatie\DataTransferObject;
 
 use ReflectionProperty;
 
+/**
+ * @internal
+ */
 class Property extends ReflectionProperty
 {
     /** @var array */
@@ -24,33 +27,25 @@ class Property extends ReflectionProperty
     /** @var bool */
     protected $isNullable = false;
 
-    /** @var bool */
-    protected $isInitialised = false;
-
     /** @var array */
     protected $types = [];
 
     /** @var array */
     protected $arrayTypes = [];
 
-    /** @var array[] */
-    protected static $cache = [];
-
-    public static function fromReflection(DataTransferObject $valueObject, ReflectionProperty $reflectionProperty)
+    public static function fromReflection(ReflectionProperty $reflectionProperty)
     {
-        return new self($valueObject, $reflectionProperty);
+        return new self($reflectionProperty);
     }
 
-    public function __construct(DataTransferObject $valueObject, ReflectionProperty $reflectionProperty)
+    public function __construct(ReflectionProperty $reflectionProperty)
     {
         parent::__construct($reflectionProperty->class, $reflectionProperty->getName());
-
-        $this->valueObject = $valueObject;
 
         $this->resolveTypeDefinition();
     }
 
-    public function set($value)
+    public function set(DataTransferObject $valueObject, $value)
     {
         if (is_array($value)) {
             $value = $this->shouldBeCastToCollection($value) ? $this->castCollection($value) : $this->cast($value);
@@ -60,9 +55,7 @@ class Property extends ReflectionProperty
             throw DataTransferObjectError::invalidType($this, $value);
         }
 
-        $this->isInitialised = true;
-
-        $this->valueObject->{$this->getName()} = $value;
+        $valueObject->{$this->getName()} = $value;
     }
 
     public function getTypes(): array
@@ -82,23 +75,10 @@ class Property extends ReflectionProperty
 
     protected function resolveTypeDefinition()
     {
-        if (isset(self::$cache[$this->class][$this->getName()])) {
-            $this->isNullable = self::$cache[$this->class][$this->getName()]['is_nullable'];
-            $this->hasTypeDeclaration = (
-                ! empty(self::$cache[$this->class][$this->getName()]['types'])
-                || ! empty(self::$cache[$this->class][$this->getName()]['array_types'])
-            );
-            $this->types = self::$cache[$this->class][$this->getName()]['types'] ?? [];
-            $this->arrayTypes = self::$cache[$this->class][$this->getName()]['array_types'] ?? [];
-
-            return;
-        }
-
         $docComment = $this->getDocComment();
 
         if (! $docComment) {
             $this->isNullable = true;
-            self::$cache[$this->class][$this->getName()]['is_nullable'] = $this->isNullable;
 
             return;
         }
@@ -107,7 +87,6 @@ class Property extends ReflectionProperty
 
         if (! count($matches)) {
             $this->isNullable = true;
-            self::$cache[$this->class][$this->getName()]['is_nullable'] = $this->isNullable;
 
             return;
         }
@@ -117,13 +96,9 @@ class Property extends ReflectionProperty
         $varDocComment = end($matches);
 
         $this->types = explode('|', $varDocComment);
-        self::$cache[$this->class][$this->getName()]['types'] = $this->types;
-
         $this->arrayTypes = str_replace('[]', '', $this->types);
-        self::$cache[$this->class][$this->getName()]['array_types'] = $this->arrayTypes;
 
         $this->isNullable = strpos($varDocComment, 'null') !== false;
-        self::$cache[$this->class][$this->getName()]['is_nullable'] = $this->isNullable;
     }
 
     protected function isValidType($value): bool
--- a/src/Property.php
+++ b/src/Property.php
@@ -84,7 +84,10 @@ class Property extends ReflectionProperty
     {
         if (isset(self::$cache[$this->class][$this->getName()])) {
             $this->isNullable = self::$cache[$this->class][$this->getName()]['is_nullable'];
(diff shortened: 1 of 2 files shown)


Gummibeer:
There is a deeper issue because of the binding of DTOs and the Properties. I will work further on this.

koomai:
@Gummibeer  thanks for this PR.

I naively assumed that caching the properties in https://github.com/spatie/data-transfer-object/blob/master/src/DataTransferObject.php#L32 would suffice, like this:

```
if (! isset(self::$cache[static::class])) {
            self::$cache[static::class] = $this->getPublicProperties($class);
        }

$properties = self::$cache[static::class];
```

Did you find any issues with this approach?

Gummibeer:
The issue is that it creates the extended property which has a binding to `$this`.
https://github.com/spatie/data-transfer-object/blob/a572b7d0e464ecdf05f03d6f61cc7b334961ca41/src/DataTransferObject.php#L143

That's why I cache it in multiple layers.

koomai:
Thoughts @brendt?

Caching significantly speeds up bulk processing.

Gummibeer:
During testing my experience is/was that the caching pays out after the second instance of the same DTO. The first is some microseconds slower because of the new checks and static vars. So someone with a million single instance DTOs will loose performance. Everyone who uses DTOs multiple times during the same runtime will benefit.

brendt:
Looking good, I'll take a close look at the code later this week and I'll merge it in!

src/Property.php
```diff
@@ -79,10 +82,23 @@ public function isNullable(): bool
...
+        if (isset(self::$cache[$this->class][$this->getName()])) {
+            $this->isNullable = self::$cache[$this->class][$this->getName()]['is_nullable'];
+            $this->hasTypeDeclaration = (
+                ! empty(self::$cache[$this->class][$this->getName()]['types'])
+                || ! empty(self::$cache[$this->class][$this->getName()]['array_types'])
+            );
+            $this->types = self::$cache[$this->class][$this->getName()]['types'] ?? [];
+            $this->arrayTypes = self::$cache[$this->class][$this->getName()]['array_types'] ?? [];
+
+            return;
```
brendt: Move this to a separate protected method `resolveFromCache` or `restoreFromCache`, maybe even another name?

src/Property.php
```diff
@@ -79,10 +82,23 @@ public function isNullable(): bool
...
+        }
+
         $docComment = $this->getDocComment();
 
         if (! $docComment) {
             $this->isNullable = true;
+            self::$cache[$this->class][$this->getName()]['is_nullable'] = $this->isNullable;
```
brendt: Why not store the class as a whole instead of its properties separately?

Gummibeer: I have an idea using clone magic method. Let me check this way. Could be that it will get a lot easier.

brendt: Nice!

Gummibeer: I have removed the coupling between the property class and the DTO instance. This way we don't have to clone the properties and only cache them. The DTO instance is now passed to the `set()` method. I've also marked the property class as `@internal` to prevent the user from using it.
https://github.com/spatie/data-transfer-object/pull/66/commits/30566493cde2672f999af5209d536ade3f5e8705

src/Property.php
```diff
@@ -24,30 +27,25 @@ class Property extends ReflectionProperty
     /** @var bool */
     protected $isNullable = false;
 
-    /** @var bool */
-    protected $isInitialised = false;
```
Gummibeer: This was only set but never used and is protected. So I think that it isn't needed anymore!?

brendt: You're right, that's an artifact from the past.

koomai:
Hey @brendt, checking to see if this PR is ready to be merged?

brendt:
Oh for some reason I didn't receive a notification about @Gummibeer 's message.

I'd like to discuss this some more though. Because we statically cache the properties on the DTO object, we need to make sure that it sets the right values on the right object. Tom has done this by removing the `Property::$valueObject` field, and instead always passing an DTO to it when calling `set`:  

```php
public function set(DataTransferObject $valueObject, $value)
{
    // …

    $valueObject->{$this->getName()} = $value;
}
```

While this solution works, semantically it's not correct anymore: a property belongs to an object, though now we need to pass that object to the property because it's loaded from cache. 

I actually had to re-read the code several times before it became clear to me what's actually happening. Now I realise caching is hard, so that's no blame on your part Tom 😉

Looking at Tom's solution, another idea came to mind, one that I think will be much more clear for future maintainers (such as myself): what if we cached the property, but unset its `valueObject` field upon caching. Loading a property from cache means we need to clone and hydrate it with our current object. 

This is roughly what that would look like:

```php
protected function getPublicProperties(ReflectionClass $class): array
{
    $properties = [];

    foreach ($class->getProperties(ReflectionProperty::IS_PUBLIC) as $reflectionProperty) {
        $properties[$reflectionProperty->getName()] = $this->getProperty($reflectionProperty);
    }

    return $properties;
}

protected function getProperty(ReflectionProperty $property): ReplectionProperty
{
    if (! isset(self::$cache[static::class][$property->getName()]) {
        self::$cache[static::class][$property->getName()] = Property::fromReflection($this, $property)->dehydrate(); // unset the reference to the DTO
    }

    $property = clone self::$cache[static::class][$property->getName()];

    $property->hydrate($this);

    return $property;

    return $property;
}
```

And property would do a check before setting values:

```php
public function set($value)
{
    if (!$this->valueObject) {
        throw new PropertyError::notHydrated($this);
    }

    // …
}
```

What are your opinions? I'm sorry to be picky about this, but I want to keep the codebase readable and maintainable in the long run. I'm perfectly fine adding this myself if Tom isn't able to spend any more time on it, let me know!

Gummibeer:
Cloning isn't a solution. Reflection are unclonable. So as long as the property class is or extends a reflection one this isn't a solution. That's the major reason why I've dropped this idea.

brendt:
Would it be possible to only cache the type information on the property class, instead of caching the property as a whole?

Gummibeer:
How about renaming the property class to validator and let it only handle the type validation.
The `set()` method only validates and does `$dto->$property = $value` that would be possible in the DTO itself and calling `$validator->isValidPropertyValue($dto, $property, $value)`.
This would remove the need of extending the reflection and get in conflicts with the naming and structure of the `set()` method.

brendt:
I think that's a good approach. But we need to make a separate PR for it. Are you up for it or will I take a look at it?

Gummibeer:
Will do it next week. 😊

Gummibeer:
@brendt before I start - would this be an acceptable class signature?
```php
namespace Spatie\DataTransferObject;

class Validator
{
    public function isValidPropertyValue(DataTransferObject $dto, string $property, $value): bool
    {

    }
}
```

Or should I change something for the primary method? I would also introduce a `Caster` which handles the current cast logic:
```php
if (is_array($value)) {
    $value = $this->shouldBeCastToCollection($value) ? $this->castCollection($value) : $this->cast($value);
}
```

brendt:
Let's open another issue to discuss this!

brendt:
#73

brendt:
Please see https://github.com/spatie/data-transfer-object/pull/77/commits/206ce5365ef8cc0ad275fa0517e67b42d212f268

Gummibeer:
Thanks for taking care of this! <3

PR #67, opened 2019-10-04
Added ability to use aliases and classes without FQCN

webdevium:


brendt:
Hi, thanks for the effort you put into this! We already considered adding this functionality: https://github.com/spatie/data-transfer-object/issues/13 and https://github.com/spatie/data-transfer-object/pull/14

Unfortunately, correctly parsing use statements is not easy. You have to take care of multiple imports in one, aliases, etc; and PHP unfortunately doesn't have reflection support for it. The truth is that we don't want to maintain such a complex parser within this package.

It's because parsing namespace imports is so hard that most coding styles and IDEs actually recommend to always use the FQDN in docblocks. I would have liked to see this differently, but as long as PHP doesn't support it via reflection, we're not going to add this in the package.

I'm sorry to close this one, like I said I appreciate the work you put into this.

PR #72, opened 2019-11-15
Automated native_function_invocation fixes

draco2003:
Reproduce with:
php php-cs-fixer --rules=native_function_invocation fix ./ --allow-risky=yes
(php-cs-fixer from : https://cs.symfony.com/)

Details:
https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/

### DIFF
--- a/src/Arr.php
+++ b/src/Arr.php
@@ -10,7 +10,7 @@ class Arr
 {
     public static function only($array, $keys): array
     {
-        return array_intersect_key($array, array_flip((array) $keys));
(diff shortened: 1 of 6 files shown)


brendt:
While I appreciate the effort, we don't want to be concerned with micro optimisations like these. We're talking about a different of nanoseconds. Those numbers will never make a real life difference compared to optimising for example IO operations. 

The difference is just too small and we don't want to think about it. If anything, PHP should fix this performance issue in the core itself, userland devs don't need to be concerned about it.

Issue #34, opened 2018-11-08
[SUGGESTION] Ability to ignore but preserve unknown properties


mfn:
If you want to use this library to encapsulate API results which are very big but you only need a fraction of it, you still have to "type" every field which makes it less useful.

Example: Twitter.

Twitters API results are nested up to 3 or 4 levels and contain a LOTS of fields. But usually you only need some parts of it, really.

For these cases, it would be useful if:
- the DTO would only need to type the properties which are going to be used (i.e. those need to be present)
- but still would preserve the rest (when transforming back via toArray)

Right now, any properties being present in the payload but not typed throw a `DataTransferObjectError`

Just to get a feeling, here's a mock of how a event for creating a direct message with attachment can look like. It just doens't make sense to type all of it, you only need access to small parts.
```
        [
            'event' => [
                'type' => 'message_create',
                'id' => '1',
                'created_timestamp' => '1525353641112',
                'message_create' => [
                    'target' => [
                        'recipient_id' => '2',
                    ],
                    'sender_id' => '2',
                    'message_data' => [
                        'text' => 'yolo https://t.co/4T35pp33WR',
                        'entities' => [
                            'hashtags' => [],
                            'symbols' => [],
                            'user_mentions' => [],
                            'urls' => [
                                [
                                    'url' => 'https://t.co/4T35pp33WR',
                                    'expanded_url' => 'https://twitter.com/messages/media/1',
                                    'display_url' => 'pic.twitter.com/4T35pp33WR',
                                    'indices' => [
                                        0 => 5,
                                        1 => 28,
                                    ],
                                ],
                            ],
                        ],
                        'attachment' => [
                            'type' => 'media',
                            'media' => [
                                'id' => 3,
                                'id_str' => '3',
                                'indices' => [
                                    0 => 5,
                                    1 => 28,
                                ],
                                'media_url' => 'https://ton.twitter.com/1.1/ton/data/dm/1/3/RBUHkp-4.jpg',
                                'media_url_https' => 'https://ton.twitter.com/1.1/ton/data/dm/1/3/RBUHkp-4.jpg',
                                'url' => 'https://t.co/4T35pp33WR',
                                'display_url' => 'pic.twitter.com/4T35pp33WR',
                                'expanded_url' => 'https://twitter.com/messages/media/1',
                                'type' => 'photo',
                                'sizes' => [
                                    'large' => [
                                        'w' => 100,
                                        'h' => 100,
                                        'resize' => 'fit',
[...]

PR #76, opened 2019-12-11
PoC to automatically cast DateTime et al property

mfn:
Since JSON has no dedicated date time type, all we get a string. The current best practice is to simply manually perform this, see https://github.com/spatie/data-transfer-object/issues/34#issuecomment-436909305 for an example
> ```php
> created_at' => Carbon::make($data['created_timestamp']),
> ```

But if there are a) many fields and b) they're in nested parts, this gets tedious quickly.

With this PoC we can set a static list of "makers" on `Property::$make` => the closure receives the $value and can "make" it whatever it needs:
```php
        Property::$make[DateTime::class] = function ($value) {
            return new DateTime($value);
        };
```
and then in the DTO:
```php
class Mydto extends DataTransferObject {
  /** @var \DateTime */
  property $date;
}
````
The reason for basically supporting arbitrary makers is that, although I've given the example with DateTime, which is ubiquitous, I might was well rather want a Carbon or Chronos object instead.

I didn't invest too much to come up with a perfect design, I was rather interested what a MVP would look like and what others thing about the idea how it could or could not go from here.

It is my understanding that this library is not concerned with:
- transformation of data
- casting of discrete values

and rightly so.

However in the pragmatic use cases I've seen, supporting date time is really a very useful property.

### DIFF
--- a/src/Property.php
+++ b/src/Property.php
@@ -8,6 +8,14 @@ use ReflectionProperty;
 
 class Property extends ReflectionProperty
 {
+    /**
+     * - The key is the type of the propoerty
+     * - The callable receives the value and may return the same or a different one.
+     * @var array<string,callable(mixed)>
+     */
+    public static $make = [
+    ];
+
     /** @var array */
     protected static $typeMapping = [
         'int' => 'integer',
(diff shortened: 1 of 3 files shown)


brendt:
> It is my understanding that this library is not concerned with:
> 
> transformation of data
> casting of discrete values
> 
> and rightly so.

Yeah I'm afraid I want to keep this package to the point: improving PHPs type system as long as it's relevant. As soon as PHP adds named parameter support and generics, we'll probably abandon this package. It's a good idea, but not for this package :)

PR #77, opened 2019-12-13
V2

brendt:
V2 adds

- PHP 7.4 only support
- `DTOCache` to improve performance
- Refactor `Property` to `FieldValidator` and `ValueCaster`
- Add `FlexibleDataTransferObject` which allows to ignore unknown properties


brendt:
@Gummibeer I'd like to hear your thoughts

src/DataTransferObject.php
```diff
@@ -29,25 +29,43 @@ public function __construct(array $parameters = [])
     {
         $class = new ReflectionClass(static::class);
```
mfn: IMHO this is unnecessary here.

The DTOCache lookup key is `static::class`, so you only need to on-demand create the reflection class in the cache resolver.

This `$class` is also used further down when throwing `unknownProperties`  error, but I believe you can simply use `static::class` for the name there as well.

src/DataTransferObject.php
```diff
@@ -29,25 +29,43 @@ public function __construct(array $parameters = [])
...
+                    $field,
+                    $validator->allowedTypes,
+                    $value
+                );
+            }
 
-            unset($parameters[$property->getName()]);
+            $this->{$field} = $value;
+
+            unset($parameters[$field]);
```
koomai: I think this shouldn't be necessary if `$ignoreMissing` is `true`.

brendt:
> Only thing I don't get is the 7.4 requirement. I could start using this library tomorrow if it weren't for that (7.3 here). Maybe I missed, but I don't see anything inherently in the design for V2 which would require it. Looks all just for the sake of sake of the new features to me.

@mfn I understand your point of view, though it's an opinion all of us at Spatie share: we want the community to move forward with modern PHP versions, and not get stuck with old PHP versions. On top of that: our internal policy for projects is to upgrade them as soon as the releases are stable, so it makes sense that we also want to upgrade the packages we use in these projects.

I realise not everyone can or wants to upgrade, and I figure these people can still use v1, which is very stable, for the years to come.

As a sidenote: it takes two hours, maybe three, to upgrade your local setup and production server, your business can afford a a few hours for the sake of security and stability.

src/DataTransferObject.php
brendt: Good point, that's fixed now!

koomai:
> > Only thing I don't get is the 7.4 requirement. I could start using this library tomorrow if it weren't for that (7.3 here). Maybe I missed, but I don't see anything inherently in the design for V2 which would require it. Looks all just for the sake of sake of the new features to me.
> 
> @mfn I understand your point of view, though it's an opinion all of us at Spatie share: we want the community to move forward with modern PHP versions, and not get stuck with older ones. On top of that: our internal policy for projects is to upgrade them as soon as the releases are stable, so it makes sense that we also want to upgrade the packages we use in these projects.
> 
> I realise not everyone can or wants to upgrade, and I figure these people can still use v1, which is very stable, for the years to come.
> 
> As a sidenote: it takes two hours, maybe three, to upgrade your local setup and production server, your business can afford a a few hours for the sake of security and stability.

Understand your point of view and am in general agreement. However, in some cases, the environment is not in our control (this is especially true in enterprise where infrastructure is managed by other teams). 

Would you be okay with a PR to v1 with the DTOCache and FlexibleDataTransferObject features from this PR?

brendt:
I'm fine with that, go for it.

mfn:
Same for me, the bigger the product, the slower the upgrades move. And it's not just always "your" code but also 3rd parties etc. I already knew the answer somehow, I just wanted to point out some impact.

@koomai would be great, thanks for the initiative 🤗

koomai:
@brendt  Thanks – did you want to create a separate 1.x branch to submit a PR against?

brendt:
No that's ok, you can target it on master for now. I'll rebase v2 on it afterwards. I'll make a v1 branch once v2 is officially tagged

brendt:
Tagged v2: https://github.com/spatie/data-transfer-object/releases/tag/2.0.0

PR #79, opened 2019-12-18
Extracted V2 features to V1

koomai:
This PR includes the V2 features from the PHP 7.4 PR (https://github.com/spatie/data-transfer-object/pull/77).

- DTOCache to improve performance
- Refactor Property to FieldValidator and ValueCaster
- Add FlexibleDataTransferObject which allows to ignore unknown properties

It leaves out the 7.4-only syntax and also the CI changes.


brendt:
Thanks, I'll tag it soon!

tests/PerformanceTest.php
```diff
@@ -0,0 +1,25 @@
...
+    /** @test */
+    public function test()
+    {
+        foreach (range(1, 500000) as $i) {
+            new MyDTO([
+                'i' => $i,
+            ]);
+        }
+
+        $this->markTestSucceeded();
```
mfn: So, this test :-)

I did notice it in https://github.com/spatie/data-transfer-object/pull/77/files#diff-9a4f6aa431675cae674bd160ff515ab2L17-L20 but was oblivious of its impact.

To give some numbers:
- the whole test suite runs in <500ms on my system (excluding this test)
- this particular test _by itself_ takes minutes; I did stop it after CPU starting to heat up

Question:
- does the test really make sense?
  Maybe benchmarking tests should be handled different, e.g. https://github.com/nuwave/lighthouse uses phpbench and a dedicate stage on travis for it: https://github.com/nuwave/lighthouse/blob/172faba3d9fc7da0e8e7394072c281ce7a7d7b9a/.travis.yml#L65-L69
- does it make sense to part of the default test suite?

Thanks!

brendt: Feel free to remove it!

mfn: Done, see https://github.com/spatie/data-transfer-object/pull/85 !

VaNGOgh94:
@koomai @mfn  Hey! Just a comment, a critique rather - introducing these changes you removed a protected method 

`    
protected function getPublicProperties(ReflectionClass $class): array
`
Since it was protected and the idea of this library is to extend library classes, this was a breaking change and IMO it should've been version 2.0.0 instead of 1.13.0. And it did break our product when upgrading from 1.11 to 1.13. Cheers

mfn:
@VaNGOgh94 Not sure why you pinged me (wasn't involved in this change), but I can see the problem.

The change happened here => https://github.com/spatie/data-transfer-object/commit/51db402dc725164ad099b2ef97f9a6b923653d1d

The whole code was (massively?) refactored and the method was now (afaics) assimilated within `\Spatie\DataTransferObject\DataTransferObject::getFieldValidators` in the cache resolver closure.

It _seems_ to me the fix however could be easy: re-introduce the method and call it from the closure. Should un-break the next 1.* release for you.

However I would suggest to parle with @brendt first if he's willing to accept such change 🤷‍♀️ (though I see it as non-conflicting in general)

PR #80, opened 2019-12-19
Improve extendability of DataTransferObject

mfn:
My previous PR https://github.com/spatie/data-transfer-object/pull/76 wasn't deemed in the scope of the library => fair point 👍 

Still I would like the ability to use the feature outlined in the PR in my own project.

As it stands right now, I've to override the whole constructor, slightly adjust it and make a copy of `getFieldValidators`.

With these changes, I'm able to simple override these methods and provide my own logic:
- allow returning my own ValueCaster instead of hardcoding `new ValueCaster` in the constructor
- ability to control the casting by overriding `castValue`
- change visibility of `getFieldValidators` from private to protected
  Otherwise someone who might want to use a custom constructor couldn't use the cache "as is"

Thanks!

### DIFF
--- a/src/DataTransferObject.php
+++ b/src/DataTransferObject.php
@@ -48,9 +48,7 @@ abstract class DataTransferObject
 
             $value = $parameters[$field] ?? $this->{$field} ?? null;
 
-            if (is_array($value)) {
-                $value = $valueCaster->cast($value, $validator);
-            }
+            $value = $this->castValue($valueCaster, $validator, $value);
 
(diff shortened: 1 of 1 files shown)


brendt:
Fair enough, I'll tag this soon!

brendt:
Released as 1.13.1: https://github.com/spatie/data-transfer-object/releases/tag/1.13.1

mfn:
Thank you!

PR #81, opened 2019-12-20
DataTransferObjectError::invalidType : get actual type before mutating $value for the error message

mfn:
I believe this is an error and wasn't caught so far because no tests for the actual message did exist.

In `\Spatie\DataTransferObject\DataTransferObjectError::invalidType` this code was executed:
```php
        $currentType = gettype($value);
```
but it was done so _after_ `$value` was possible overwritten like this:
```php
        if ($value === null) {
            $value = 'null';
        }

        if (is_object($value)) {
            $value = get_class($value);
        }

        if (is_array($value)) {
            $value = 'array';
        }
```
Which basically made `gettype($value);` to yield `string` in the majority of cases.

This PR fixes this.
Please also see the 2nd commit stand-alone, showing the change form the old to the new message  => https://github.com/spatie/data-transfer-object/commit/182c9251e5fb78e6a7c53d5c7f48a72a5537017e

### DIFF
--- a/tests/DataTransferObjectTest.php
+++ b/tests/DataTransferObjectTest.php
@@ -80,7 +80,7 @@ class DataTransferObjectTest extends TestCase
     public function null_is_allowed_only_if_explicitly_specified()
     {
         $this->expectException(DataTransferObjectError::class);
-        $this->expectExceptionMessageRegExp('/Invalid type: expected `class@anonymous[^:]+::foo` to be of type `string`, instead got value `null`, which is string/');
+        $this->expectExceptionMessageRegExp('/Invalid type: expected `class@anonymous[^:]+::foo` to be of type `string`, instead got value `null`, which is NULL/');
 
         new class(['foo' => null]) extends DataTransferObject {
             /** @var string */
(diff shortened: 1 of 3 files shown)


tests/DataTransferObjectTest.php
```diff
@@ -79,6 +80,7 @@ public function default_values_are_supported()
     public function null_is_allowed_only_if_explicitly_specified()
     {
         $this->expectException(DataTransferObjectError::class);
+        $this->expectExceptionMessageRegExp('/Invalid type: expected `class@anonymous[^:]+::foo` to be of type `string`, instead got value `null`, which is NULL/');
```
mfn: Before this PR, this message ended in:
> instead got value `null`, which is string

and now
> instead got value `null`, which is NULL

tests/DataTransferObjectTest.php
```diff
@@ -27,6 +27,7 @@ public function only_the_type_hinted_type_may_be_passed()
         $this->markTestSucceeded();
 
         $this->expectException(DataTransferObjectError::class);
+        $this->expectExceptionMessageRegExp('/Invalid type: expected `class@anonymous[^:]+::foo` to be of type `string`, instead got value ``, which is boolean/');
```
brendt: Why the need for regex matching? Wouldn't it be possible to do a string equals string assertion?

mfn: Unfortunately not, because they contain some PHP internal magic object ID stuff which is "random" as far as the test is concerned.

I wish I wouldn't have to do that 😄 

This only applies to tests using anonymous classes within.

mfn: _and_ it also includes:
- the local path 💥
- a (I think) `0`-byte 🤷‍♀️
  Pretty dark magic 🧙‍♂️ going on there


Here's an example how the error looks if it doesn't mach:
> `Failed asserting that exception message 'Invalid type: expected class@anonymous/Users/neo/src/data-transfer-object/tests/DataTransferObjectTest.php0x1126bc4d3::foo to be of type string, instead got value , which is boolean.' matches '/Invalid type: expected class@annymous[^:]+::foo to be of type string, instead got value "", which is boolean/'.`

brendt: I see, thanks for clarifying!

brendt:
Released as 1.13.2: https://github.com/spatie/data-transfer-object/releases/tag/1.13.2

Thanks for all the PRs!

PR #82, opened 2019-12-20
[DTO] Ignore static public properties

mfn:
I've a project where I need to use a public static property on a DTO but
it clashes because the code doesn't ignore it. But it should, as they're
not relevant for the data encapsulation of the DTO.


brendt:
Thanks! Will tag later today

brendt:
Tagged as 1.13.0: https://github.com/spatie/data-transfer-object/releases/tag/1.13.0

mfn:
Thank you!

PR #83, opened 2019-12-20
[DTO] Add arrayOf static helper

mfn:
I've a use case where I received payloads directly representing an array of
some DTO. Now, I could use _another_ DTO to encapsulate this via a nested DTO
approach, but there are cases where this is simply overkill.

This helper makes this easier:
```php
class Mydto extends DataTransferObject {
  /** @var string */
  public $name;
}

$myDtos = Mydto::arrayOf([
  ['name' => 'foo'],
  ['name' => 'bar],
]);
```


brendt:
Nice PR, thanks! Will tag later today

brendt:
Tagged as 1.13.0: https://github.com/spatie/data-transfer-object/releases/tag/1.13.0

mfn:
Thank you!

PR #86, opened 2020-01-08
fix: immutability for nested object

kaiweia:


brendt:
Thanks! Will tag later today

PR #94, opened 2020-02-17
Update README.md

Propaganistas:
As per https://github.com/laravel/framework/pull/13566#issuecomment-219420623 this readme should also reference the `input()` method instead of  `get()`.


brendt:
TIl, thanks

PR #95, opened 2020-03-04
Fix uninitialized property check

squatto:
The `$hasDefaultValue` property on `FieldValidator` was incorrectly being set to the value of `$property->isDefault()`. `isDefault()` indicates if a property was declared at compile time or created at run time - it doesn't have anything to do with whether or not the property has a default value ([PHP docs for ReflectionProperty::isDefault](https://www.php.net/manual/en/reflectionproperty.isdefault.php))

Here is a slightly modified version of the first example from the docs:

```php
<?php

class Foo {
    public $bar;
}

$o = new Foo();
$o->bar = 42;
$o->baz = 42;

$ro = new ReflectionObject($o);
var_dump($ro->getProperty('bar')->isDefault()); // bool(true)
var_dump($ro->getProperty('baz')->isDefault()); // bool(false)
```

`bar->isDefault()` is `true` because it was declared in the class (note that it _doesn't_ have a default value).

This PR modifies `$hasDefaultValue` to instead look at the default properties of the class and determine if the default value for the property is not `null`, which would indicate that it has a default value. I also modified three of the tests that should have been seeing the "uninitialized" error message, but were instead seeing the "invalidType" error message.

### DIFF
--- a/src/FieldValidator.php
+++ b/src/FieldValidator.php
@@ -39,9 +39,11 @@ class FieldValidator
 
     public static function fromReflection(ReflectionProperty $property): FieldValidator
     {
+        $defaultProperties = $property->getDeclaringClass()->getDefaultProperties();
+
         return new self(
             $property->getDocComment() ?: null,
-            $property->isDefault()
+            ($defaultProperties[$property->name] ?? null) !== null
         );
     }
 
--- a/tests/DataTransferObjectTest.php
+++ b/tests/DataTransferObjectTest.php
@@ -81,7 +81,7 @@ class DataTransferObjectTest extends TestCase
     public function null_is_allowed_only_if_explicitly_specified()
     {
         $this->expectException(DataTransferObjectError::class);
-        $this->expectExceptionMessageRegExp('/Invalid type: expected `class@anonymous[^:]+::foo` to be of type `string`, instead got value `null`, which is NULL/');
+        $this->expectExceptionMessageRegExp('/Non-nullable property `class@anonymous[^:]+::foo` has not been initialized./');
 
         new class(['foo' => null]) extends DataTransferObject {
             /** @var string */
@@ -237,7 +237,7 @@ class DataTransferObjectTest extends TestCase
     public function an_exception_is_thrown_when_property_was_not_initialised()
     {
         $this->expectException(DataTransferObjectError::class);
-        $this->expectExceptionMessageRegExp('/Invalid type: expected `class@anonymous[^:]+::foo` to be of type `string`, instead got value `null`, which is NULL/');
+        $this->expectExceptionMessageRegExp('/Non-nullable property `class@anonymous[^:]+::foo` has not been initialized./');
 
         new class([]) extends DataTransferObject {
             /** @var string */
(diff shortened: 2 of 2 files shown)


brendt:
Thanks for the PR! What if a property has a default value of `null` though? I believe it'll always be marked as not having a default value.

Maybe we should use `property_exists` ?

brendt:
Scratch everything I said, I misread the code. It looks good! I'll merge this soon

brendt:
(sorry for all the edits)

I don't think the scenario is covered when a default value is null. I think you'll need to use an `array_key_exists` call on the default properties array.

squatto:
> I don't think the scenario is covered when a default value is null. I think you'll need to use an `array_key_exists` call on the default properties array.

Correct me if I'm wrong but isn't setting the default value of a class property to `null` the same as not setting it at all? I would think that both would be considered as not having a default value.

For reference, check the [php docs on `null`](https://www.php.net/manual/en/language.types.null.php):

>A variable is considered to be null if:
>* it has been assigned the constant `NULL`.
>* it has not been set to any value yet.
>* it has been unset().

Specifically the first two points: assigned `NULL` === not set

I tested a DTO with both no default on a property and with `null` as the default:

```php
// DTO property without a default value

/** @var string|int */
public $company_id;

// $property->getDeclaringClass()->getDefaultProperties()
[
  "company_id" => null
]
```

```php
// DTO property with a default value of `null`

/** @var string|int */
public $company_id = null;

// $property->getDeclaringClass()->getDefaultProperties()
[
  "company_id" => null
]
```

The outcome of both is exactly the same, so both should be considered as not having a default value.

brendt:
As of PHP 7.4, there's an "uninitialised" state, which is what `hasDefaultValue` represented before 7.4. Since this package is aiming to mimic PHP's type system as close as possible, I think it's worth taking this into consideration.

In practice though, there's an explicit `null` check: https://github.com/spatie/data-transfer-object/blob/master/src/DataTransferObject.php#L54-L56 so your code would work, though I think that null check should be removed in the long run. 

Take the following examples:

```php
class MyDTO extends DataTransferObejct
{
    public ?int $a = null; // default value is null, 
    // and is valid to be omitted

    public int $b; // no default value provided, 
    // its value is required to be set

    /** @var int|null */
    public $c; // The package allows this, 
    // because it's a feature added before PHP-74

    public ?int $d; // However, PHP 7.4 does NOT allow to leave this value empty, 
    // you need to explicitly pass an integer or null, not leave it empty
}
``` 

Right now, our type checks work like `$c`, which isn't in line with what PHP 7.4 does. 

I think `array_key_exists` is the right way to go, if we want to support the proper type checks in v2.

squatto:
Aaaah, gotcha, I didn't realize that typed properties came along with a new uninitialized state! Thank you for pointing that out.

I will modify my code to use `array_key_exists` for `$hasDefaultValue`. However, the problem still exists where it isn't properly determining if something is uninitialized. In the latest release, `DataTransferObjectError::uninitialized()` errors never actually occur because non-nullable properties that aren't set in the DTO constructor cause `DataTransferObjectError::invalidType()` errors, as you can see in this test:

https://github.com/spatie/data-transfer-object/blob/98f90a291fb519efd521efc1569b9821ce354dd3/tests/DataTransferObjectTest.php#L237-L246

I'm going to need to upgrade my local machine to 7.4 in order to test the new uninitialized state. I will get that done and then update this PR with my new changes.

Thanks for your help!

brendt:
FYI, as of PHP 7.4, `ReflectionProperty` also has an `isInitialized` method: https://www.php.net/manual/en/reflectionproperty.isinitialized.php

PR #96, opened 2020-03-04
Add properties to exceptions

squatto:
Because the same `DataTransferObjectError` exception is thrown for every possible error, it is difficult to properly handle the specific error. This PR adds additional contextual information about the specific error to the exception object and exposes that information through various getter methods. The getter methods are explained in the readme.

Another way of doing this would be to create separate exceptions for each error type that extend the base `DataTransferObjectError` class. I can refactor to that if preferred.

Thanks!

### DIFF
--- a/src/DataTransferObjectError.php
+++ b/src/DataTransferObjectError.php
@@ -8,11 +8,23 @@ use TypeError;
 
 class DataTransferObjectError extends TypeError
 {
+    use HasErrorProperties;
+
+    public function __construct($message, $properties = [])
+    {
+        parent::__construct($message);
+        $this->setProperties($properties);
+    }
+
     public static function unknownProperties(array $properties, string $className): DataTransferObjectError
     {
         $propertyNames = implode('`, `', $properties);
 
-        return new self("Public properties `{$propertyNames}` not found on {$className}");
+        return new self("Public properties `{$propertyNames}` not found on {$className}", [
+            'error'      => __FUNCTION__,
+            'class'      => $className,
+            'properties' => $properties,
+        ]);
     }
 
     public static function invalidType(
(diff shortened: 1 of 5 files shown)


squatto:
Here is an example implementation of this change in which I am modifying the exception message based on the type and throwing a different exception that's specific to the process I'm building. I'm modifying the exception message to be more non-developer-friendly because it is being logged and displayed to an end user. Without the changes in this PR, I would have to use regular expressions to extract text from the exception message, which creates a hard dependency on this package never changing the error messages.

```php
private function throwParsingException(DataTransferObjectError $exception)
{
    $message = $exception->getMessage();

    if ($exception->getError() === 'invalidType') {
        $message = "The field '{$exception->getProperty()}' " .
                   "is expected to be of type '" . implode(', ', $exception->getExpectedTypes()) . "', " .
                   "but the value of '{$exception->getValue()}' " .
                   "has a type of '{$exception->getType()}'";
    }

    if ($exception->getError() === 'uninitialized') {
        $message = "The required field '{$exception->getProperty()}' was not provided or does not have a value";
    }

    throw new CrmDataFileParsingException(
        $message,
        $this->dataFile,
        $exception->getProperty(),
        $this->lineNumber,
        $this->line,
        $exception
    );
}
```

tests/DataTransferObjectTest.php
```diff
@@ -36,6 +36,24 @@ public function only_the_type_hinted_type_may_be_passed()
...
 
+    /** @test */
+    public function invalid_types_set_exception_properties()
+    {
+        try {
+            new class(['foo' => 'value']) extends DataTransferObject {
+                /** @var boolean */
+                public $foo;
+            };
+        } catch (DataTransferObjectError $error) {
```
mfn: Just a heads-up: if the library code changes and throws no exception, it might go unnoticed here (might still show up as a test having no assertions though).

I suggest a different approach (for this and the other tests):
```php
$error = null;
try {
  …
} catch (…) {
  // Deliberately empty
}
$this->asssertInstanceOf(DataTransferObjectError::class, $error);
$this->assertEquals(…)
```

src/DataTransferObjectError.php
```diff
@@ -8,11 +8,23 @@
...
+        $this->setProperties($properties);
+    }
+
     public static function unknownProperties(array $properties, string $className): DataTransferObjectError
     {
         $propertyNames = implode('`, `', $properties);
 
-        return new self("Public properties `{$propertyNames}` not found on {$className}");
+        return new self("Public properties `{$propertyNames}` not found on {$className}", [
+            'error'      => __FUNCTION__,
```
mfn: 🤔
Wouldn't it make more sense to have dedicated exception for all this (for backwards compat extend from `DataTransferObjectError`) ?

brendt:
> Another way of doing this would be to create separate exceptions for each error type that extend the base DataTransferObjectError class. I can refactor to that if preferred.

I think that's the way to go, use the type system to its full extent!

src/DataTransferObjectError.php
squatto: I definitely think that's the best approach as well. See [this comment](https://github.com/spatie/data-transfer-object/pull/96#issuecomment-595150363) - I'm going to refactor my changes to add separate exceptions for each error type.

squatto:
> > Another way of doing this would be to create separate exceptions for each error type that extend the base DataTransferObjectError class. I can refactor to that if preferred.
> 
> I think that's the way to go, use the type system to its full extent!

OK perfect, coming right up!

tests/DataTransferObjectTest.php
squatto: I used an existing test as the template for my new tests. You're definitely right - this creates the possibility of hiding assertions if the throws change. Great catch, thank you!

PR #99, opened 2020-03-11
readme: update collection examples

mfn:
- I introduced arrayOf some time ago and it turns out, it's perfect with for the collection
- `@method` can be used, IDEs (PhpStorm) can still properly autocomplete this.

Since the `Collection` is supposed to be a small enhancement for type safety, I thought making it less boiler-platey might be welcome


brendt:
Nice!

PR #101, opened 2020-03-28
Support basic array<…> and array<…,…> syntax

mfn:
Problem: I can't use `array<int,string>` or `array<string,string>` (for e.g. static analyzers) because I always get an exception like this:
```
Spatie\DataTransferObject\DataTransferObjectError : Invalid type: expected `DTO::Property` to be of type `array<string`, instead got value `array`, which is array.
```
This PR is an attempt to fix this, see the tests.

In case of violations, the error message isn't very helpful, it will be something like:
```
Invalid type: expected `DTO::Property` to be of type `array<string,string>`, instead got value `array`, which is array.
```
But that's a limitation of the design of the `\Spatie\DataTransferObject\FieldValidator::isValidType`: it can only return `true`/`false` but would need more specific details here.

I don't think the implementation is "super cool", but gets the job done. It's basically the first brain-dump approach to get it working. Would appreciate feedback to improve this!

### DIFF
--- a/src/FieldValidator.php
+++ b/src/FieldValidator.php
@@ -33,7 +33,7 @@ class FieldValidator
     /** @var array */
     public $allowedTypes = [];
 
-    /** @var array */
+    /** @var array<string|array{0:string,1:string}> */
     public $allowedArrayTypes = [];
 
 
@@ -48,7 +48,7 @@ class FieldValidator
     public function __construct(?string $docComment = null, bool $hasDefaultValue = false)
     {
         preg_match(
-            '/@var ((?:(?:[\w?|\\\\<>])+(?:\[])?)+)/',
+            '/@var ((?:(?:[\w?|\\\\<>,])+(?:\[])?)+)/',
             $docComment ?? '',
             $matches
         );
@@ -108,9 +108,25 @@ class FieldValidator
         return $value instanceof $type || gettype($value) === $type;
     }
 
-    private function assertValidArrayTypes(string $type, $collection): bool
+    /**
+     * @param string|array $type
+     * @param iterable $collection
+     * @return bool
+     */
+    private function assertValidArrayTypes($type, $collection): bool
     {
-        foreach ($collection as $value) {
+        foreach ($collection as $key => $value) {
+            if (is_array($type)) {
+                if (! $this->assertValidType($type[0], $key)) {
+                    return false;
+                }
+                if (! $this->assertValidType($type[1], $value)) {
+                    return false;
+                }
+
+                continue;
+            }
+
             if (! $this->assertValidType($type, $value)) {
                 return false;
             }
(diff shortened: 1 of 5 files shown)


src/FieldValidator.php
```diff
@@ -48,7 +48,7 @@ public static function fromReflection(ReflectionProperty $property): FieldValida
     public function __construct(?string $docComment = null, bool $hasDefaultValue = false)
     {
         preg_match(
-            '/@var ((?:(?:[\w?|\\\\<>])+(?:\[])?)+)/',
+            '/@var ((?:(?:[\w?|\\\\<>,])+(?:\[])?)+)/',
```
mfn: Added `,` so it matches it in `array<…,…>`

src/FieldValidator.php
```diff
@@ -33,7 +33,7 @@ class FieldValidator
     /** @var array */
     public $allowedTypes = [];
 
-    /** @var array */
+    /** @var array<string|array{0:string,1:string}> */
```
mfn: It's now either a string or a string tuple holding the key and the value type.

src/FieldValidator.php
```diff
@@ -171,16 +187,29 @@ function (string $type) {
                     return str_replace(['iterable<', '>'], ['', ''], $type);
                 }
 
+                if (strpos($type, 'array<') !== false) {
+                    $arrayTypes = explode(',', str_replace(['array<', '>'], ['', ''], $type));
```
mfn: This `explode` gets the ball rolling to support this feature.

mfn:
ping @brendt 🤗

src/FieldValidator.php
```diff
@@ -108,9 +108,25 @@ private function assertValidType(string $type, $value): bool
...
+    /**
+     * @param string|array $type
+     * @param iterable $collection
+     * @return bool
+     */
+    private function assertValidArrayTypes($type, $collection): bool
     {
-        foreach ($collection as $value) {
+        foreach ($collection as $key => $value) {
+            if (is_array($type)) {
```
brendt, line 119: I'm not sure about passing the type as an array. It's all rather implicit, and you're never actually certain about what's in that array. 

I think we've arrived at a point where a dedicated `FieldType` with some concrete implementations are in order.

A `FieldType` would be able to validate a given value. I think we'd need three or four such implementations: `ScalarType`, `ObjectType`, `GenericType` and `ArrayKeyValueType`. 

What's your opinion?

mfn: Agree :)

Didn't wanted to invest too much upfront, was rather a quick PoC.

Well draft the PR and look at the feedback!

mfn: For now I just replaced the `string|array` with a single `FieldType`, this is not yet your idea about `FieldType`.

Doing that would result in an even more massive refactoring and I'm not sure if I should go in that direction.

mfn:
A lack a bit of vision how I could go about this and not enough to be creative => closing!

brendt:
No problem, I might look into this myself soon! Thanks for all contributions!

PR #103, opened 2020-04-24
[TRAVIS] Run tests against PHP 7.4

mfn:
While working on something locally I noticed the tests breaking, but not because of my code: I'm using 7.4 and the exception message checks failed because the internal `class@anonymous` format changed slightly.

Adapted the regex in a way it works for 7.4 and earlier


brendt:
I've just tagged v2, which doesn't use travis anymore :)

mfn:
@brendt I guess you pulled the trigger too fast 😄 

Master is failing because in 7.4 the exception message changed which we've tests for which now fail.

Cherry-pick https://github.com/spatie/data-transfer-object/pull/103/commits/184f086cb6aa483eccea8710b09f2317996d5fcd into master and it should be good!

brendt:
Thanks! I was actually going to look into these today, but this makes it more easy 👍

PR #100, opened 2020-03-11
feat: interpret php 7.4 types

skyrpex:
Closes https://github.com/spatie/data-transfer-object/issues/98

PR #105, opened 2020-05-06
Support typed properties

adam-prickett:
With PR adds support for typed properties in the PHP 7.4 without the need for docblock comments.

There are limitations, in that typed arrays are not yet supported, however primitives, nullable types, objects and nested DTOs are all supported.

Gauging interest in the approach here, so if this is agreeable I'll updated README too.

### DIFF
--- a/src/FieldValidator.php
+++ b/src/FieldValidator.php
@@ -4,8 +4,8 @@ declare(strict_types=1);
 
 namespace Spatie\DataTransferObject;
 
-use ReflectionProperty;
 use ReflectionNamedType;
+use ReflectionProperty;
 
 class FieldValidator
 {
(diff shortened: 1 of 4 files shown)


brendt:
I'm actually working on this myself, I'll try to submit a PR soon so that we can compare our approaches, because they do differ.

mfn:
Isn't there also https://github.com/spatie/data-transfer-object/pull/100 trying the same? 🤔

brendt:
Yes, it's clear that everyone want proper support for nested DTOs with php 7.4 types 😁 I'll fix it right now!

brendt:
Closed in favour of https://github.com/spatie/data-transfer-object/pull/106

PR #109, opened 2020-05-13
Fix: Validator for typed 7.4 properties

ABSAhmad:
The reflection types don't seem to be being passed to the TypeMapping which leads to wrong results in the validator because it uses `gettype` which has a different result than the reflection types.


brendt:
Thanks!

brendt:
Released as 2.2.1: https://github.com/spatie/data-transfer-object/releases/tag/2.2.1

Issue #111, opened 2020-05-13
Support for self/static


jpnut:
Currently any references to `self` or `static` as type hints or type declarations do not work. When normalising the type, it is converted to a string. That means when we call `assertValidType`, we check if `$value` is an instance of `'self'` or of `'static'`. (To be absolutely clear, I'm referring to `static` here not in the sense of a static property but as a reference to the called class).

I did a small investigation, the changes required to implement this would be small but they would necessitate an api change for the `FieldValidator` abstract class (and thus the implementations).

For example, one solution could be: when normalising the type, first check to see if the type is `'self'` or `'static'`. For `self`, we should return the class which declares the property. For `static` we should return the class we are validating (hence the api change).

Some thoughts: supporting `self` alone is very low priority since `self` can always be replaced by the class name. `static`, on the other hand, cannot be replaced in general. The way I see it, if you decided to support `static`, it would then make sense to support `self` too. 

As an example usage of `static`, consider the following:

```php
abstract class Foo extends DataTransferObject
{
    /**
     * @var static|null
     */
    public $child;
}

class Bar extends Foo
{
    public string $baz;
}

$bar = new Bar([
    'child' => [
        'baz' => '...'
    ],
    'baz' => '...',
]);
```

In the above, `child` is resolved to an instance of `Bar`.

If this is something the package should support, I'd be happy to submit a PR.

Thanks for your work on this package, it's a pleasure to use! 👍

brendt:
How would you know what concrete class should be used as the `static` one though? 

Take the following example:

```php
class A extends DataTransferObject
{
    /** @var static|null */
    public $staticProp;
}

class StaticB extends A 
{
    public string $name;
}

class StaticC extends A 
{
    public string $name;
}

$a = new A([
    'staticProp' =>  [
        'name' => 'test' // Is this StaticB or StaticC?
    ],
]);
```

There's no way to know this, that's the nature of static.

jpnut:
In your example, surely `staticProp` should be resolved to an instance of A? It would only be resolved to `StaticB` or `StaticC` if those classes were constructed explicitly.

Consider for a moment an example using the same structure as yours above, but with 

```php
class A extends DataTransferObject
{
    /** @var \A|null */
    public $a;
}
```

instead. Then the following are all valid:

```php
$a = new A([
    'a' =>  [], // This is resolved to an instance of A
]);

$b = new A([
    'a' => new StaticB([
        'name' => 'test'
    ])
]);

[...]

PR #112, opened 2020-05-14
Add support for self & static property types

jpnut:
see #111

This adds support for `static` and `self` type declarations. `self` accepts any instance or extension of the class in which it is _defined_. `static` accepts any instance or extension of the class in which it is _called_. When resolving from an array, `self` is always resolved to an instance of the class in which it is defined. `static` is resolved to an instance of the class in which it is called.

Unfortunately, whilst the change in functionality is small, the code base change is large since implementations of `FieldValidator` must now _always_ have information about the property and class instance.

I'm happy to make any changes upon request 👍

### DIFF
--- a/src/DataTransferObject.php
+++ b/src/DataTransferObject.php
@@ -184,7 +184,7 @@ abstract class DataTransferObject
 
                 $field = $reflectionProperty->getName();
 
-                $properties[$field] = FieldValidator::fromReflection($reflectionProperty);
+                $properties[$field] = FieldValidator::fromReflection($reflectionProperty, $class);
             }
 
             return $properties;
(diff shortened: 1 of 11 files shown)


brendt:
Sorry it took me so long to get back to this. Honestly I'm still debating whether we want to add this or not. You convinced me in the issue it was worth looking into, but those are a lot of changes for a feature that — I think — will almost never be used.

I understand why you'd want  `self`, but `static` just seems so weird in the context of DTOs. 

I appreciate the effort, but I'm still not convinced, sorry; especially since I'm the one who has to maintain this code for the years to come 😁

We can discuss the matter further, but I think I'll be closing the PR for now.

jpnut:
I completely understand, no worries at all. I was originally planning on using this library as part of another project where this functionality was quite important, but as time went on I realised it wasn't entirely necessary.

Rest assured: my time wasn't wasted as I was able to use this idea in a different project, and I fully appreciate your reasons for not adding this to the library!

brendt:
Hi! I really appreciate that input, thanks! Best of luck with the project!

PR #117, opened 2020-05-31
Cast nested collections

### LINKED ISSUE #116: Automatic casting of nested array DTOs doesn't work with DataTransferObjectCollection
```php
class TagData extends DataTransferObject
{
   public string $name;
}

class TagDataCollection extends DataTransferObjectCollection
{
    public function current(): TagData
    {
        return parent::current();
    }
}

{
   public TagDataCollection $tags;
}

$postData = new PostData([
    'tags' => [
        ['name' => 'foo'],
        ['name' => 'bar']
    ]
]);
```

Would you be open for a PR?

morrislaptop:
A bit more code changes than I would of liked, but here goes:

* Will cast to `DataTransferObjectCollection` objects instead of an array if typed by the parent `DataTransferObject`
* Supports docblock definition of the collection class (will use the first if multiple types are defined)
* Foundation for PHP 8 Union types
* Throws an exception if type is not specified for a collection class

Fixes #116


brendt:
Thanks, I'm going to review and merge this tomorrow :) Could you take a quick look at the merge conflicts with master still?

aidan-casey:
@morrislaptop Regarding the questions in #116, the type should also be inferred from `offsetGet()`, not `offsetSet()`. Otherwise you are right on.

morrislaptop:
@aidan-casey ah yep, fixed my typo! 

@brendt Do you want to implement support for `offsetGet()`. Will need to support docblocks and type hints and decide what to do if they are different? Throw an exception or treat as a union?

morrislaptop:
Otherwise merge conflicts resolved and tests passing :)

aidan-casey:
Definitely something for @brendt to answer, but an early return on the first one retrieved might be a good approach. Most people probably don't set `offsetGet()` given it is not mentioned in the docs and it'd just be weird to have them return two different things.

morrislaptop:
@brendt were you still intending to merge this?

milewski:
I tried this PR.. It worked as expected

mfn (changes requested):

src/PropertyFieldValidator.php
```diff
@@ -59,22 +58,20 @@ private function resolveIsMixedArray(ReflectionProperty $property): bool
     private function resolveAllowedTypes(ReflectionProperty $property): array
     {
         // We cast to array to support future union types in PHP 8
-        $types = [$property->getType()];
+        $types = [$property->getType()
+            ? $property->getType()->getName()
+            : null, ];
```
mfn, line 63: Gnah, auto styling #fail 😅

morrislaptop:
Haha I'm sure brendt will fix when merging

brendt:
Lost track of this one! I'll merge it this week :)

morrislaptop:
@brendt bump 👊

src/DocblockFieldValidator.php
```diff
@@ -85,7 +88,7 @@ private function resolveAllowedTypes(string $definition): array
 
     private function resolveAllowedArrayTypes(string $definition): array
     {
-        return $this->normaliseTypes(...array_map(
+        return $this->normaliseTypes(...(new RecursiveIteratorIterator(new RecursiveArrayIterator(array_map(
```
brendt, line 91: Can you explain the need for these iterators?

src/FieldValidator.php
```diff
@@ -148,4 +149,63 @@ private function assertValidType(string $type, $value): bool
     {
         return $value instanceof $type || gettype($value) === $type;
     }
+
+    protected function resolveAllowedArrayTypesFromCollection(string $type): array
+    {
+        if (! $type || ! class_exists($type) || ! is_subclass_of($type, DataTransferObjectCollection::class)) {
```
brendt, line 155: Wouldn't `! is_subclass_of($type, DataTransferObjectCollection::class)` suffice? It'd also return false for the other conditions I believe.

src/FieldValidator.php
```diff
@@ -148,4 +149,63 @@ private function assertValidType(string $type, $value): bool
...
+
+        $currentReturnType = $class->getMethod('current')->getReturnType();
+
+        // We cast to array to support future union types in PHP 8
+        $currentReturnTypes = [];
+        if ($currentReturnType) {
+            $currentReturnTypes[] = $currentReturnType->getName();
+        }
+
+        $docblockReturnTypes = $class->getDocComment()
```
brendt, line 169: I don't think I want to support the `/** @method … current */` notation on the class itself. That's extra overhead for something that's always possible by overriding the method itself no?

src/PropertyFieldValidator.php
```diff
@@ -59,22 +58,20 @@ private function resolveIsMixedArray(ReflectionProperty $property): bool
...
 
         return $this->normaliseTypes(...$types);
     }
 
-    private function normaliseTypes(?ReflectionType ...$types): array
```
brendt, line 67: Why was normaliseTypes removed?

src/ValueCaster.php
```diff
@@ -32,7 +32,7 @@ public function castValue($value, array $allowedTypes)
         return new $castTo($value);
     }
 
-    public function castCollection($values, array $allowedArrayTypes)
+    public function castCollection($values, array $allowedArrayTypes, string $collectionClass = null)
```
brendt, line 35: I'm not sure I like that this method's signature is growing. Any alternatives?

brendt (changes requested):

src/FieldValidator.php
morrislaptop: Correct, and if we remove this then it will vastly simplify the PR, going to attempt this now

src/ValueCaster.php
morrislaptop: Refactored

niektenhoopen (changes requested):

README.md
```diff
@@ -245,7 +245,7 @@ Alternatively you can also use a phpdoc for this:
 use Spatie\DataTransferObject\DataTransferObjectCollection;
 
 /**
- * @method PostData current
+ * @method \App\DTOs\PostData current
```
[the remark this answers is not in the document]
aidan-casey: @niektenhoopen I am pretty sure this is fine. It is used as an example in the README and necessary to achieve the auto completion mentioned. Note how these namespaces are used elsewhere in the existing README.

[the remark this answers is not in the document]
niektenhoopen: Ah, I couldn't find it and I have now figured out why. In the existing readme it's `App\DataTransferObjects\PostData` instead of `App\DTOs\PostData`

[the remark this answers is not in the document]
aidan-casey: Yeah, event though it is longer it may be nice to have it be `App\DataTransferObjects\PostData` just to be consistent. I wouldn't let that keep the feature from coming through though. 👍🏼

niektenhoopen, line 248: Let's see if this works. That would made it a one-click-change.
```suggestion
 * @method \App\DataTransferObjects\PostData current
```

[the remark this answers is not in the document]
niektenhoopen: Let's see if this works. It would make it a one-click-change:
```suggestion
 * @method \App\DataTransferObjects\PostData current
```

bilfeldt:
@morrislaptop This is an awesome addition to the package.

This addition cast nested DTOs directly to collections as stated in your example:

```php
class TagData extends DataTransferObject
{
   public string $name;
}

class TagDataCollection extends DataTransferObjectCollection
{
    public function current(): TagData
    {
        return parent::current();
    }
}

{
   public TagDataCollection $tags;
}

$postData = new PostData([
    'tags' => [
        ['name' => 'foo'],
        ['name' => 'bar']
    ]
]);
```

But how would you construct the `TagDataCollection` directly from a nested array?

```php
$tagDataCollection = new TagDataCollection([
    ['name' => 'foo'],
    ['name' => 'bar'],
]);

$tagDataCollection->current(); // TypeError: Return value of TagDataCollection::current() must be an instance of TagData, array returned
```

In the current version this is only possible using:
```php
$tagDataCollection = new TagDataCollection([
    new TagData(['name' => 'foo']),
    new TagData(['name' => 'bar']),
]);
```

Being able to convert a `DataCollection` to and from nested arrays would be the missing link in adding this as a custom cast in Laravel: https://github.com/jessarcher/laravel-castable-data-transfer-object/issues/4

Ideally we should be able to do something like:

```php
$data = [
    'tags' => [
        ['name' => 'foo'],
        ['name' => 'bar']
    ]
];

$postData = new PostData(data);
$postData->toArray() == $data; // true - This is possible with this PR

$tagDataCollection = new TagDataCollection($data['tags']); - This is not possible correct?
$tagDataCollection->toArray() == $data['tags']; // true - This is already implemented!
```

aidan-casey:
@bilfeldt - To clarify, you are simply talking about adding the casting ability to the collection as well, so that if it is expecting a certain type it will cast to that type automatically. Correct?

@brendt would have to speak into this more than I can, but the collections are only a basic implementation to support handling multiple DTOs out of the box. They are kept lightweight and it seems best to keep them that way by avoiding adding casters to them as well.

That said, in Laravel you can always _not_ use the Spatie DTO Collection and instead extend the Laravel Collection built-in. From there you could add a caster directly to it for this functionality. 👍🏼

basepack:
@brendt can you merge this in yet, or am I missing tomething. I do get the same error `// TypeError: Return value of TagDataCollection::current() must be an instance of TagData, array returned�` as @bilfeldt at the moment though.

brendt (approved):

brendt:
@morrislaptop anything else you want to change before merging? Could you also rebase this branch on master?

morrislaptop:
Nope - rebasing now

brendt:
It's ok like so, thanks!

brendt:
Released as 2.7.0: https://github.com/spatie/data-transfer-object/releases/tag/2.7.0

PR #126, opened 2020-07-14
README.md update to match PHP 7.4

### LINKED ISSUE #124: README.md update to match PHP 7.4
Hey!

First off - lovely package (as always)! 💯

Would be really nice if README.md could be updated with the current preferred standart on how to use DTOs after 7.4 was released and the package's minimum required PHP version was bumped to 7.4. Currently on the README.md it still says: 

> When PHP 7.4 introduces typed properties, you'll be able to simply remove the doc blocks and type the properties with the new, built-in syntax.

aidan-casey:
Updated README.md to use PHP 7.4 syntax. Closes #124.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,8 @@ You can install the package via composer:
 composer require spatie/data-transfer-object
 ```
 
+* **Note**: This package requires PHP 7.4 so it can take full advantage of type casting in PHP.
+
 ## Support us
 
 Learn how to create a package like this one, by watching our premium video course:
@@ -112,9 +114,7 @@ class PostData extends DataTransferObject
 }
 ```
 
-Since PHP 7.4 introduced typed properties, you can use the new, built-in syntax. Their values will be validated against the given type and a `TypeError` will be thrown if the value does not comply with it.
-
-* **NOTE:** Watch out for the upcoming release of PHP 8. More typing features are on the way!
+When defining typed properties, you can take advantage of the built-in types supported by PHP. These properties will be validated against the given type and a `TypeError` will be thrown if the value does not comply with it.
 
 ```php
 use App\Models\Author;
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ class PostData extends DataTransferObject
     public string $property;
 
     /**
-     * Classes with their FQCN:
+     * Imported class or fully qualified class name:
      */
     public Author $property;
 
--- a/README.md
+++ b/README.md
@@ -305,19 +305,29 @@ $postData = new PostData([
 
 ### Automatic casting of nested array DTOs
 
-Similarly to above, nested array DTOs will automatically be cast.
+Similarly to above, nested array DTOs will automatically be cast. For example, we can define the following DTO:
 
 ```php
+namespace App\DataTransferObjects;
+
 use Spatie\DataTransferObject\DataTransferObject;
 
 class TagData extends DataTransferObject
 {
    public string $name;
 }
+```
+
+By referencing this object in our `PostData` DTO, a `TagData` DTO will be automatically cast.
+
+```php
+namespace App\DataTransferObjects;
+
+use Spatie\DataTransferObject\DataTransferObject;
 
 class PostData extends DataTransferObject
 {
-    /** @var \TagData[] */
+    /** @var \App\DataTransferObjects\TagData[] */
    public $tags;
 }
 ```
@@ -332,7 +342,7 @@ $postData = new PostData([
     ]
 ]);
 ```
-**Attention**: Remember, for nested type casting to work, your Docblock definition needs to be a Fully Qualified Class Name (`\App\DTOs\TagData[]` instead of `TagData[]` and a use statement at the top).
+**Attention**: Remember, for nested type casting to work, your Docblock definition needs to be a Fully Qualified Class Name (`\App\DataTransferObjects\TagData[]` instead of `TagData[]` and a use statement at the top).
 
 ### Immutability
 
--- a/README.md
+++ b/README.md
@@ -228,7 +228,7 @@ If you're working with collections of DTOs, you probably want auto completion an
 This package adds a simple collection implementation, which you can extend from.
 
 ```php
-use PostData;
+use App\DataTransferObjects\PostData;
 use Spatie\DataTransferObject\DataTransferObjectCollection;
 
 class PostCollection extends DataTransferObjectCollection
@@ -267,7 +267,7 @@ $postCollection[0]-> // … and also here.
 Of course you're free to implement your own static constructors:
 
 ```php
-use PostData;
+use App\DataTransferObjects\PostData;
 use Spatie\DataTransferObject\DataTransferObjectCollection;
 
 class PostCollection extends DataTransferObjectCollection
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -62,14 +62,11 @@ We did our best to keep the syntax and overhead as little as possible:
...
-    public $title;
+    public string $title;
     
-    /** @var string */
-    public $body;
+    public string $body;
     
-    /** @var \Author */
-    public $author;
+    public \Author $author;
```
mfn, line 69: I think the examples should "import" the real classes when used here; and maybe this should be documented that this works.

The important thing to remember is that for phpdoc you CANNOT use the imported version, but this restriction doesn't apply to _actual_ types as the type is already correctly resolved by the PHP engine itself when reflections are used.

README.md
```diff
@@ -109,10 +106,41 @@ class PostData extends DataTransferObject
 }
 ```
 
-By adding doc blocks to our properties, their values will be validated against the given type; 
-and a `TypeError` will be thrown if the value doesn't comply with the given type.
+Since PHP 7.4 introduced typed properties, you can use the new, built-in syntax. Their values will be validated against the given type and a `TypeError` will be thrown if the value does not comply with it.
 
-Here are the possible ways of declaring types:
+* **NOTE:** Watch out for the upcoming release of PHP 8. More typing features are on the way!
```
mfn, line 111: Are there? 🤔

README.md
aidan-casey: Yeah I am not opposed to that. Looks better and more fluent with the rest of the doc.

README.md
aidan-casey: Yup! So far we have union and mixed types coming. This will solve our inability to replicated those doc block types documented.

https://php.watch/versions/8.0/union-types
https://php.watch/versions/8.0/mixed-type

README.md
aidan-casey: Latest commit provides an update that resolves this.

README.md
```diff
@@ -60,16 +60,15 @@ This package allows you to create data transfer object definitions, classes, whi
 We did our best to keep the syntax and overhead as little as possible:
 
 ```php
+use \App\Models\Author;
```
mfn, line 63: Imports are usually written without leading `\` as they're always qualified from the "root namespace 😄 Applies to the other examples too:
```suggestion
use App\Models\Author;
```

README.md
```diff
@@ -180,14 +217,13 @@ class PostData extends DataTransferObject
...
-When PHP 7.4 introduces typed properties, you'll be able to simply remove the doc blocks and type the properties with the new, built-in syntax.
-
 ### Working with collections
 
 If you're working with collections of DTOs, you probably want auto completion and proper type validation on your collections too.
 This package adds a simple collection implementation, which you can extend from.
 
 ```php
+use \PostData;
 use \Spatie\DataTransferObject\DataTransferObjectCollection;
```
mfn, line 227: Yeah this one should be fixed too, no one writes it that way
```suggestion
use Spatie\DataTransferObject\DataTransferObjectCollection;
```

README.md
```diff
@@ -109,10 +110,46 @@ class PostData extends DataTransferObject
 }
 ```
 
-By adding doc blocks to our properties, their values will be validated against the given type; 
-and a `TypeError` will be thrown if the value doesn't comply with the given type.
+Since PHP 7.4 introduced typed properties, you can use the new, built-in syntax. Their values will be validated against the given type and a `TypeError` will be thrown if the value does not comply with it.
```
mfn, line 113: Re-reading the readme in its entirety, it strikes me that this should be already at the beginning of the readme where "Installation" is. _Usually_ a readme has something like "requirements" (this doens't) but I think the first thing mentioned is that it requires 7.4.

Following this, not sure if this sentence, explaining that 7.4 has typed properties, makes even sense because.. it's like news to use "now" but in fact you.

Maybe add the aforementioned to the beginning at the installation and then here something like:
> You can take full advantage of PHP 7.4 and its typed properties. Their values …

README.md
aidan-casey: Yeah, I am aware. Since that was how it was written in the original, I did not want to be too opinionated.

README.md
aidan-casey: Yeah I agree with you. I am not sure about throwing it under the installation section though. As a developer reading through, seeing built-in typing vs phpdoc typing back to back helps me see some of the feature differences.

Perhaps we need to revise the existing text to something along the lines of:

> When defining typed properties, you can take advantage of the built-in types supported by PHP. These properties will be validated against the given type and a `TypeError` will be thrown if the value does not comply with it.

And add this to the installation section:

> * **Note**: This package requires PHP 7.4 so it can take full advantage of type casting in PHP.

README.md
aidan-casey: Going to resolve this conversation in the duplicate thread.

README.md
aidan-casey: Took a look at other Spatie packages and found that this scenario of prefixing a use statement with a backslash is unique. I'll make another commit to remove this.

aidan-casey: Resolved by eb0f5ad45c211bff4488468673d4147035bfb802

README.md
mfn: > so it can take full advantage

I was about to say that's not correct but (7.4 is the minimum)… of course versions <2 do support < PHP 7.4 so 🤷‍♀️

aidan-casey: @mfn They are also tagged and should have a different README.

aidan-casey: Resolved by 70c516805f63e94297a3baab34f8771904be2ee5

README.md
```diff
@@ -109,12 +114,49 @@ class PostData extends DataTransferObject
...
-Here are the possible ways of declaring types:
+class PostData extends DataTransferObject
+{
+    /**
+     * Built in types:
+     */
+    public string $property;
+
+    /**
+     * Classes with their FQCN:
```
jdreesen, line 132: This should be something like "`use`d classes", no?

aidan-casey: Whoops. Yeah, I am inclined to make that:

> Imported class or fully qualified class name:

Will commit shortly. Who knew it would be so much work just to update a README! 😄

aidan-casey: Resolved by 7fd5a5ab8c1514e3bb81eddceb9bd1b79a43be22

README.md
```diff
@@ -285,7 +332,7 @@ $postData = new PostData([
     ]
 ]);
 ```
-**Attention**: For nested type casting to work your Docblock definition needs to be a Fully Qualified Class Name (`\App\DTOs\TagData[]` instead of `TagData[]` and an use statement at the top)
+**Attention**: Remember, for nested type casting to work, your Docblock definition needs to be a Fully Qualified Class Name (`\App\DTOs\TagData[]` instead of `TagData[]` and a use statement at the top).
```
mfn, line 335: Btw, the example it refers to (above, not touched in this PR [yet 😄 ]) uses `… @var \TagData[]` and not `\App\…\DataData`; seems like an existing typo there and should probably be updated, too?

aidan-casey: Basically we need to update our example app directory structure. We can do that. 😄

brendt:
Thanks!

rihardssceredins:
Thanks guys, sorry I didn't come up with PR myself for this!

PR #130, opened 2020-08-04
Collect invalid type errors instead of stopping at the first

mfn:
## Summary
I deal with a lot of bigger DTOs, lots of properties. Invalid payloads
happen all the time but it's annoying to only always get the first
invalid type reported when they're more to wait.

This PR will not change single invalid type reporting but will support
multiple lines:
```
Spatie\DataTransferObject\DataTransferObjectError : The following invalid types were encountered:
expected `App\Services\DTO\Article::title` to be of type `string`, instead got value `null`, which is NULL
expected `App\Services\DTO\Article::created_at` to be of type `Carbon\CarbonImmutable`, instead got value `null`, which is NULL
```

Before finishing this (see below for TODO) I first want to know if @brendt wants to accept this?

### TODO
- [x] Move `\Spatie\DataTransferObject\DataTransferObjectError::invalidTypeMessage` out of the exception
- [x] Add tests for new behaviour

### DIFF
--- a/tests/DataTransferObjectTest.php
+++ b/tests/DataTransferObjectTest.php
@@ -501,4 +501,33 @@ class DataTransferObjectTest extends TestCase
 
         $this->assertSame($expected, $object->all());
     }
+
+    /** @test */
+    public function multiple_required_properties_are_reported()
+    {
+        $exception = null;
+
+        try {
+            new class() extends DataTransferObject {
+                public string $i_am_required;
+                public int $so_am_i;
+            };
+        } catch (DataTransferObjectError $exception) {
+            // Deliberately empty
+        }
+
+        $this->assertInstanceOf(DataTransferObjectError::class, $exception);
+        
+        $actual = $exception->getMessage();
+        $actual = preg_replace('/anonymous.*:/', 'anonymous:', $actual);
+
+        $expected = <<<EXPECTED
+The following invalid types were encountered:
+expected `class@anonymous:i_am_required` to be of type `string`, instead got value `null`, which is NULL
+expected `class@anonymous:so_am_i` to be of type `integer`, instead got value `null`, which is NULL
+
+EXPECTED;
+
+        $this->assertSame($expected, $actual);
+    }
 }
--- a/src/DataTransferObject.php
+++ b/src/DataTransferObject.php
@@ -46,6 +46,9 @@ abstract class DataTransferObject
 
         $valueCaster = $this->getValueCaster();
 
+        /** string[] */
+        $invalidTypes = [];
+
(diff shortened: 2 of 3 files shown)


mfn:
Friendly ping @brendt 🏖️

brendt:
Hi! Thanks for the effort and sorry for the late reply, I was on holiday. I'm not sure what to think of this… PHP's type system also stops at the first errors, and you know how I want this package to stay as close to PHP's type system as possible… 🤔

mfn:
> I want this package to stay as close to PHP's type system as possible

Ah, ok; I was not quite aware of _that_, note taken!

Not sure what to argument here.

Having done a major refactoring lately in a project (many many DTOs, many many properties, payloads changing left and right) I did temporarily commited these changes and I can only say... _it was a blessing to see ALL type errors instead of just the first one_.

So my only argument is that it's a super DX enhancement and I definitely want to "keep" this. In the worst case I've to override the `DataTransferObject` constructor in my project which I rather not would do… (at this point I can almost just copy the library in my source tree 😅).

Let me know if the DX argument is persuading enough 😄

tests/DataTransferObjectTest.php
```diff
@@ -501,4 +501,33 @@ public function ignore_static_public_properties_when_serializing()
...
+                public int $so_am_i;
+            };
+        } catch (DataTransferObjectError $exception) {
+            // Deliberately empty
+        }
+
+        $this->assertInstanceOf(DataTransferObjectError::class, $exception);
+        
+        $actual = $exception->getMessage();
+        $actual = preg_replace('/anonymous.*:/', 'anonymous:', $actual);
```
mfn, line 522: I decided not to use `expectExceptionMessageRegExp` because the regex for the muliltine message was crazy to formulate and hardly readable and I figured: clarity trumps so I just catch it => massage the message to remove the environment dependent paths => then simply match it 1:1

brendt:
It's a very persuading argument. Let me sleep a night on this. 

Also sorry for the late reply, I'm still rather busy the coming next two weeks…

mfn:
Ping 😼 🙏

brendt:
@mfn I had to manually merge it due to a merge conflict: https://github.com/spatie/data-transfer-object/commit/8648e5ea67e9d9504294e2709038995750c82205

So I'll close this PR, but it is merged!

brendt:
Released as 2.5.0: https://github.com/spatie/data-transfer-object/releases/tag/2.5.0

mfn:
> So I'll close this PR, but it is merged!

Wohoo, thanks!

Issue #129, opened 2020-07-20
phpstan and checkUninitializedProperties


mfn:
Today phpstan https://github.com/phpstan/phpstan/releases/tag/0.12.33 was released, bringing us:
> Check uninitialized typed properties that aren't set in the constructor (https://github.com/phpstan/phpstan-src/commit/a1eb1f5b65b62019b1d4653f9438c7591c3379a3), https://github.com/phpstan/phpstan/issues/2984
> -   Off by default, needs checkUninitializedProperties: true.

This bombs on any DTO with require fields as they're detected as being un-initialized 😅

> Class App\DTO\SomeDTO has an uninitialized property $id. Give it default value or assign it in the constructor.

For this library this static analysis doesn't matter as a DTO with required fields will never incorrectly created.

I'm creating this issue to discuss what best practices we can provide here:
- tell people to just ignore it (e.g. baseline)
  Could be annoying when DTO changes
- come up with a configuration to ignore this particular rule for anything extending from `\Spatie\DataTransferObject\DataTransferObject`
- custom phpstan rule to deal with this 🤔

mfn:
@ondrejmirtes may you can provide a recommendation? This is thanks to your latest https://github.com/phpstan/phpstan/releases/tag/0.12.33 release bringing us checkUninitializedProperties 🎉

This library here works like this, you provide:
```php
class FancyDTO extends \Spatie\DataTransferObject\DataTransferObject
{
  public int $id;
}
```
and use it like this:
```php
$dto = new FancyDTO(['id' => 123']);
```

`DataTransferObject` itself will not construct the object if such properties ("required attributes") are not provided in the payload to the constructor, e.g.
```php
$dto = new FancyDTO([]);
```
will throw an exception.

As in: this library will never create objects where public uninitialized properties can be present.

I'm using this in a  (Laravel) project where I could already fix some real bugs (thank you 🙏 ) because I enabled the rule but I use a lot of such DTOs and thus get a LOT of errors.

I can add them to the baseline, but DTOs get changed/added and this creates lots of noise.

Is it possible to ignore a rule for a certain "pattern" of file? 🤔

Or is it possible to write a rule to ignore other rules 😅

Or is it possible to mark descendants of `DataTransferObject` save for this, similarly as you marked phpunits `setUp` as constructor?

Thanks for any pointers 😄

mfn:
[...]

PR #135, opened 2020-08-16
Add extension for PHPStan to support checkUninitializedProperties

mfn:
## Summary
Please read the discussion in https://github.com/spatie/data-transfer-object/issues/129 first.

I decided to go ahead and suggest a simple inclusion of the PHPStan extension for now instead of separate repo to avoid the overhead of managing a dedicate package for a few lines of code.

The idea is:
- the extension installed
- phpstan is suggested for composer
- end users can enable the extension manually with this neon configuration
  ```neon
  services:
    -
      class: Spatie\DataTransferObject\PHPstan\PropertiesAreAlwaysInitializedExtension
      tags:
        - phpstan.properties.readWriteExtension
  ```

Tests are added with a dedicated workflow:
- installs phpstan (just the latest version for now)
- runs a positive and a negative test, i.e. ensures that activating/absence of the extension works correctly.

Ad github action: I decided now to cache composer but use `composer:v2` instead. In my tests, due to parallel downloads, it was as fast as the v1 with cache or very close.

### Links
- Fixes https://github.com/spatie/data-transfer-object/issues/129

### DIFF
--- a/.github/workflows/phpstan-extension.yml
+++ b/.github/workflows/phpstan-extension.yml
@@ -1,0 +1,31 @@
+name: Test PHPStan extension
+
+on: [push]
+
+jobs:
+  phpstan-extension:
+    runs-on: ubuntu-latest
+    env:
+        COMPOSER_NO_INTERACTION: 1
+
+    steps:
+      -   name: Checkout code
+          uses: actions/checkout@v2
+
+      -   name: Setup PHP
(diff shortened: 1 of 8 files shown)


brendt:
Awesome! Great addition!

brendt:
Released as 2.3.0: https://github.com/spatie/data-transfer-object/releases/tag/2.3.0

PR #138, opened 2020-08-28
Convert already nicely documented regex to true multi-line regex

mfn:
I mean, it's already there, right? :-)

### DIFF
--- a/src/DocblockFieldValidator.php
+++ b/src/DocblockFieldValidator.php
@@ -6,25 +6,26 @@ namespace Spatie\DataTransferObject;
 
 class DocblockFieldValidator extends FieldValidator
 {
-    /*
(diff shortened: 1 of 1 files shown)


brendt:
See, now I'm confused: I tried exactly this, and all tests failed, which is why I moved the docs to the comments 🤔

brendt:
Oh, is that what the `x` modifier does?

jdreesen:
It is, see `PCRE_EXTENDED` here: https://www.php.net/manual/en/reference.pcre.pattern.modifiers.php

brendt:
TIL! 

Thanks!

mfn:
❤️ 

Yeah, I too use like once in 5 years or so…

PR #140, opened 2020-09-02
Use fromRequest method if exists in nested DTOs.

abdulrahman19:
Hi,

I open this PR because I had issue with nested DTOs, that I can't format income data using `fromRequest`. ex:
```php
class Category extends DataTransferObject
{
    /** @var string */
    public $name;

    /** @var \Bla\To\Product */
    public $product;
}

class Product extends DataTransferObject
{
    /** @var string */
    public $name;

    /** @var \Carbon\Carbon */
    public $expiration;

    public static function fromRequest($data): self
    {
        return new self([
            'name'      => $data['name'],
           'expiration' => Carbon::parse($data['expiration'])
        ]);
    }
}

$category_dto = new Category([
    'name' => 'bla',
    'product' => [
        'name' => 'product',
        'expiration' => '2020-09-02 12:22' // this will give an error
    ],
]);
```

for this issue I opened this PR to solve this problem, now if it find `fromRequest` it will use it, instead of create new object and pass data to constructor directly

### DIFF
--- a/src/ValueCaster.php
+++ b/src/ValueCaster.php
@@ -29,7 +29,9 @@ class ValueCaster
             return $value;
         }
 
-        return new $castTo($value);
+        return (method_exists($castTo, 'fromRequest'))
+            ? $castTo::fromRequest($value)
+            : new $castTo($value);
     }
 
     public function castCollection($values, array $allowedArrayTypes)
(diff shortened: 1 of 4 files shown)


brendt:
I'm afraid this functionality is out of scope for this package. You're free to extend the base DTO class and override the functionality in your project.

PR #141, opened 2020-09-09
Override protected attributes

devemio:
Hi guys,

If I understand correctly, I can't add some field names to my DTO class to use them when parsing some data. These are protected fields that are declared in the parent `DataTransferObject` class. For example, `ignoreMissing`, `exceptKeys`, and `onlyKeys`.

```php
use Spatie\DataTransferObject\DataTransferObject;

{
    public int $ignoreMissing;
    
    public bool $onlyKeys;
}
```

I also noticed that the `except()` and `only()` methods are mainly used in combination with the `toArray()` method, since these methods only affect the list of keys when exporting to an array, perhaps we can transfer such methods to a separate class and also use a chain of method calls.

Thank you for your attention.

### DIFF
--- a/src/DataTransferObject.php
+++ b/src/DataTransferObject.php
@@ -9,8 +9,6 @@ use ReflectionProperty;
 
 abstract class DataTransferObject
 {
-    protected bool $ignoreMissing = false;
-
     protected array $exceptKeys = [];
 
     protected array $onlyKeys = [];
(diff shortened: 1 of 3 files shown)


brendt:
At first I didn't think I'd merge this. But giving it a little more thought it's actually a nice feature 🤔

Unfortunately we'd need to tag a new major release, since we're changing the signature of existing public methods, and we're also removing the properties — which technically could be set in DTO implementations to override the defaults. 

So I'll revisit this one when we plan on tagging a new major release, that will probably happen when PHP 8 arrives.

mfn (approved):
Will definitely break my use of the library 😏 but that change truly makes sense to me 👍 

Let's just sure we mention this prominently in, wherever the right place is, so people are ware of it.

PR #149, opened 2020-11-16
Change DataTransferObject's field validator/cast functions to static

ekisu:
The motivation behind this is that, if one wants to use the validators in some other way aside from instantiating the DTO with valid data (for example, writing a custom validator function to check which fields are valid and whatnot), currently it's not possible, as the functions are only available after instantiating the class, meaning you have to use valid data in the first place.

Changing those functions to static makes them available for this usecase.

### DIFF
--- a/src/DataTransferObject.php
+++ b/src/DataTransferObject.php
@@ -42,9 +42,9 @@ abstract class DataTransferObject
 
     public function __construct(array $parameters = [])
     {
-        $validators = $this->getFieldValidators();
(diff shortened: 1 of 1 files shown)


brendt:
I've always thought of validators and casts as an internal implementation detail of DTOs. Opening them up to the outside world doesn't seem like the best idea, since suddenly these things can do way more than they were intended to do.

However, I'm going to work on a new version with the release of PHP 8, which will probably focus much more on casts and validators, so I might revisit this one later.

PR #150, opened 2020-11-17
Adds Data Transfer Object Factory to the README

### LINKED ISSUE #148: Data Transfer Object Factory - external tool?
Hey there! We built [this](https://github.com/Anteris-Dev/data-transfer-object-factory) package that generates DTOs based on their definition. It utilizes ReflectionClass and and phpDocumentor's doc block parser to determine a property's type and faker to generate a value for that property. If a property does not have a type assigned, it will have a random type generated.

Wondering if this might be a candidate for the external tools section. If so, I am happy to PR, but otherwise that's okay too!

aidan-casey:
Closes #148 by adding a reference to the Data Transfer Object Factory package in the external tools section.


brendt:
Very nice, thanks! I'll try this out later!

PR #152, opened 2020-11-25
add support on php 8

Nielsvanpach:


mfn:
Yikes!

This is annoying, as all the failed tests are due to the changed error message.

I was never sure if this was a good idea (I started to add them), but we had some issues regarding them which went unnoticed.

_Probably_ this can be fixed with adjusting the regex, let me know if you need help.

PS: not sure about 99% of the other changes though? 🤔
I would expect the composer.json php bump and the matrix addition + the test fixes at this point, but the rest?

brendt:
@Nielsvanpach let's look at this together in a call today :)

brendt:
The solution is pretty simple: only test for part of the exception message instead of the whole. I'll merge this one later today.

PR #156, opened 2020-12-22
Any object that has a callable toArray() should be parsed as array

niektenhoopen:
Not sure if this change follows the logic of this package but I like the package to be more flexible with accepting (nested) objects that should be parsed as arrays.

With this change an attribute can also be for example a Laravel Collection.

I would like to get some feedback and I'm happy to make changes if necessary.

### DIFF
--- a/src/DataTransferObject.php
+++ b/src/DataTransferObject.php
@@ -154,10 +154,7 @@ abstract class DataTransferObject
     protected function parseArray(array $array): array
     {
         foreach ($array as $key => $value) {
-            if (
-                $value instanceof DataTransferObject
-                || $value instanceof DataTransferObjectCollection
-            ) {
+            if (is_object($value) && is_callable([$value, 'toArray'])) {
                 $array[$key] = $value->toArray();
 
                 continue;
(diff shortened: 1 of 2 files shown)


bilfeldt:
@niektenhoopen Shouldn't this also be added to `DataTransferObjectCollection@ toArray ` here: https://github.com/spatie/data-transfer-object/blob/64555c085364b567701ff627ef442ff1b2d42757/src/DataTransferObjectCollection.php#L74-L90

niektenhoopen:
> @niektenhoopen Shouldn't this also be added to `DataTransferObjectCollection@toArray` here:
> 
> https://github.com/spatie/data-transfer-object/blob/64555c085364b567701ff627ef442ff1b2d42757/src/DataTransferObjectCollection.php#L74-L90

Probably. It's not something I use but it does make sense for consistency. I have added the change. I hope @brendt also can give his opinion because the current implementation is quite opinionated and I can not decide for him if my way is also his preferred way.

brendt:
We're not going to add this functionality in this package, sorry. It's too opinionated, as @niektenhoopen says. For example: what about `__toArray` or `asArray`, should we also support those notations? 

My suggestion is to extend from the DTO collection class and add the functionality on a project-level basis.

Issue #163, opened 2021-02-04
Wrong return type in data-transfer-object/src/FieldValidator.php


rusowyler:
Hi guys in getCurrentReturnTypesFromDocblock (line 185) you expect and array as return value:
 
`    private function getCurrentReturnTypesFromDocblock(string $definition): array
`

and on line 197 you may return null:

` if (! $type) {
            return null;
        }`

so if if there is doc block and no math to the preg_match it will throw a exception!

You should probably replace: return null; 
with: return [];

Hope it helps!

brendt:
Feel free to submit a PR fixing this issue!

brendt:
Oh, you already did; thanks!

PR #164, opened 2021-02-04
Fix incorrect return type

rusowyler:
Return type should be an array (with data or empty) and not null.

See https://github.com/spatie/data-transfer-object/issues/163


mfn (approved):
👍

brendt:
Released: https://github.com/spatie/data-transfer-object/releases/tag/2.8.1

Issue #167, opened 2021-02-16
Automatic casting of nullable nested array DTOs


pikant:
I work with the Shopify API, which has the special feature that you can use a `fields` parameter with every GET request to specify which fields should be included in the response. This procedure is intended to save data volume.

This results in the specialty that I have to make all fields of my DTOs nullable. This also works so far, only with nested arrays (see https://github.com/spatie/data-transfer-object#automatic-casting-of-nested-array-dtos) it doesn't seem to work.

I am getting the following error:
```
   Spatie\DataTransferObject\DataTransferObjectError 

  Invalid type: expected `App\Domain\Shopify\DataTransferObjects\Order\Order::line_items` to be of type `\App\Domain\Shopify\DataTransferObjects\Order\LineItems[]`, instead got value `null`, which is NULL..

  at vendor/spatie/data-transfer-object/src/DataTransferObjectError.php:24
     20▕         $msg = count($invalidTypes) > 1
     21▕             ? "The following invalid types were encountered:\n" . implode("\n", $invalidTypes) . "\n"
     22▕             : "Invalid type: {$invalidTypes[0]}.";
     23▕ 
  ➜  24▕         throw new self($msg);
     25▕     }
     26▕ 
     27▕     public static function invalidTypeMessage(
     28▕         string $class,
```

Here is an excerpt from the DTO:
```php
class Order extends FlexibleDataTransferObject
{
    public? int $ id;

    // ...

    / ** @var \ App \ Domain \ Shopify \ DataTransferObjects \ Order \ LineItems [] $ line_items * /
    public? array $ line_items;
```

I'm about to create a PR with some tests that reproduces the behavior. I will link this PR here.

mfn:
Is it a problem with contradicting phpdoc? phpdoc does not say it's nullable and I think it has higher precedence if present?

pikant:
Thanks for the quick answer. Indeed I tried a trailing ' | null' in the phpdoc. However, I just realized, that it only works without the whitespaces. Is this a phpdoc constraint?

pikant:
I also updated this change in my PR and renamed it also. Only in case you want to include the tests in the code: https://github.com/spatie/data-transfer-object/pull/168

This issue here can be closed.

Many Thanks!

mfn:
> However, I just realized, that it only works without the whitespaces

Never ever have I see these whitespaces phpdoc in all the years. I actually thought this was a freak copy paste mistake on your part:
```
[...]

PR #168, opened 2021-02-16
Added tests for automatic casting of nullable nested array DTOs

pikant:
See https://github.com/spatie/data-transfer-object/issues/167


brendt:
Thanks!

PR #175, opened 2021-04-02
Update FUNDING.yml

patinthehat:
This PR updates `FUNDING.yml` to use the same format/data as other Spatie packages.


brendt:
Thanks!

PR #176, opened 2021-04-02
Fix cast with null

bastien-phi:
Currently, a DTO like 
```php
class ComplexDtoWithNullableProperty extends DataTransferObject
{
    public string $name;

    public ?BasicDto $other;
}
```
cannot be created with
```php
new ComplexDtoWithNullableProperty(name: 'a', other: null)
```
because we are having a
```
TypeError: Only arrays and Traversables can be unpacked

vendor/spatie/data-transfer-object/src/Casters/DataTransferObjectCaster.php:17
vendor/spatie/data-transfer-object/src/Reflection/DataTransferObjectProperty.php:41
```

This PR fixes the issue applying the cast only when given value is not null.

### DIFF
--- a/src/Reflection/DataTransferObjectProperty.php
+++ b/src/Reflection/DataTransferObjectProperty.php
@@ -37,7 +37,7 @@ class DataTransferObjectProperty
 
     public function setValue(mixed $value): void
     {
-        if ($this->caster) {
+        if ($this->caster && $value !== null) {
             $value = $this->caster->cast($value);
         }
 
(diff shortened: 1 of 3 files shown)


brendt:
Aah, yes, thanks for sending a PR, gonna tag immediately!

brendt:
Done! https://github.com/spatie/data-transfer-object/releases/tag/3.0.1

bastien-phi:
Wait a minute, I'm just sending another PR

bastien-phi:
You're too quick for me ! :joy:

brendt:
I don't mind many tags ;)

PR #177, opened 2021-04-02
Give dto to caster

bastien-phi:
It is sometimes convenient to give an already created nested DTO to the parent DTO.

For example, if you want to map an api response which looks like 
```json
{  
  "product": "the product",
  "value": {
    "cents": 100,
    "sign": "€"      
  }
}
```
to
```php
class Product extends DataTransferObject {

    public string $label;
    public string $price;
    
    public static function fromApi(array $api) 
    {
        return new self([
            'label' => $api['product'], 
            'price' => Price::fromApi($api['value'])
        ]);
    }
}

class Price extends DataTransferObject {

    public int $amount;
    public string $currency;


   public static function fromApi(array $api) 
   {
        return new self([
            'amount' => $api['cents'],
            'currency' => $api['sign']
       ]);
    }
}
```
It is currently not possible as DataTransferObject will try to cast it as well.

~~This fix may be implemented upstream, directly in DataTransfertObjectProperty~~ Not sure about that

### DIFF
--- a/src/Casters/DataTransferObjectCaster.php
+++ b/src/Casters/DataTransferObjectCaster.php
@@ -14,6 +14,10 @@ class DataTransferObjectCaster implements Caster
 
     public function cast(mixed $value): DataTransferObject
     {
+        if ($value instanceof $this->className) {
+            return $value;
+        }
+
         return new $this->className(...$value);
     }
 }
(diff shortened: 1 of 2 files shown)


brendt:
Good one as well! Great!

brendt:
Tagged: https://github.com/spatie/data-transfer-object/releases/tag/3.0.2 :)

brendt:
I tagged the wrong commit… hang on

brendt:
Tagged https://github.com/spatie/data-transfer-object/releases/tag/3.0.2

bastien-phi:
Thanks !

PR #178, opened 2021-04-02
Fix when nested DTO have casted field

bastien-phi:
Not yet sure how to fix this but it looks like I found a little issue.

I will dig into it but feel free to add the fix to this PR !

### DIFF
--- a/src/Reflection/DataTransferObjectProperty.php
+++ b/src/Reflection/DataTransferObjectProperty.php
@@ -6,7 +6,6 @@ use JetBrains\PhpStorm\Immutable;
 use ReflectionAttribute;
 use ReflectionClass;
 use ReflectionProperty;
-use Spatie\DataTransferObject\Arr;
 use Spatie\DataTransferObject\Attributes\CastWith;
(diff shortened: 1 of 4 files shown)


bastien-phi:
Looks like I found a fix.

Removing the `dd()` is required here (not sure why it was here BTW)

src/Reflection/DataTransferObjectProperty.php
```diff
@@ -125,6 +126,6 @@ private function resolveCasterFromDefaults(): ?Caster
             }
         }
 
-        dd($defaultCastAttributes);
+        return null;
```
mfn, line 129: 😄

bastien-phi:
Hi @brendt !

Sorry for pinging you. Is there something wrong with this PR ? I hope this is the correct way to fix the issue !

brendt:
> not sure why it was here BTW 

That was my bad, thanks for removing it. Looks good, thanks for fixing!

brendt:
Released as 3.0.3: https://github.com/spatie/data-transfer-object/releases/tag/3.0.3

bastien-phi:
Thanks for merging !

PR #183, opened 2021-04-08
*Very* minor typo fix...

telkins:
I'm assuming it should be `item`, which makes sense to me, but perhaps I've missed something.... 🤔


brendt:
You're correct, thanks for the PR!

PR #184, opened 2021-04-08
Add jetbrains attributes package

vlakarados:
As the code is filled with JetBrains helper attributes like `#[Immutable]` now, it's best to include the package in `require-dev`. This will solve issues with other IDE's and self-sufficient.

https://packagist.org/packages/jetbrains/phpstorm-attributes
https://github.com/JetBrains/phpstorm-attributes


brendt:
Thanks!

PR #185, opened 2021-04-09
Fix union type casting

cappuc:
Right now properties with a union type are not supported because the `resolveCasterFromType` method in `DataTransferObjectProperty` class throw an exception in the `$type->getName()` call (The `getName` method does not exists in `ReflectionUnionType`.

I don't know if the solution I implemented is the best solution but in my tests it works.


brendt:
Thanks, I already had this ready in a separate branch, but I'm happy to merge your implementation :) Will tag later today

brendt:
FYI I removed the null branch in the match block, that should never happen.

brendt:
Tagged as 3.0.4

PR #188, opened 2021-04-10
Feature Idea: Uncast / Normalize / Serialize objects

morrislaptop:
This library is fantastic for mapping incoming data to objects but it could also be great for outgoing data as well.

I've added a sample test to show you what I mean. 

Would this be within the scope of this project? Happy to work on this further if so :)

### DIFF
--- a/src/Attributes/DefaultUncast.php
+++ b/src/Attributes/DefaultUncast.php
@@ -1,0 +1,58 @@
+<?php
+
+namespace Spatie\DataTransferObject\Attributes;
+
+use Attribute;
+use ReflectionProperty;
+use ReflectionNamedType;
+use ReflectionUnionType;
+use JetBrains\PhpStorm\Immutable;
+use Spatie\DataTransferObject\Caster;
+use Spatie\DataTransferObject\Uncaster;
+use Spatie\DataTransferObject\Exceptions\InvalidUncasterClass;
+
+#[Attribute(Attribute::TARGET_CLASS)]
+class DefaultUncast
+{
+    public function __construct(
+        #[Immutable]
+        private string $targetClass,
+        #[Immutable]
+        private string $uncasterClass,
+    ) {
+        if (! class_implements($this->uncasterClass, Uncaster::class)) {
+            throw new InvalidUncasterClass($this->uncasterClass);
+        }
+    }
+
+    public function accepts(ReflectionProperty $property): bool
+    {
+        $type = $property->getType();
+
+        /** @var \ReflectionNamedType[]|null $types */
+        $types = match ($type::class) {
+            ReflectionNamedType::class => [$type],
+            ReflectionUnionType::class => $type->getTypes(),
+            default => null,
+        };
+
+        if (! $types) {
+            return false;
+        }
+
+        foreach ($types as $type) {
+            if ($type->getName() !== $this->targetClass) {
+                continue;
+            }
+
+            return true;
+        }
+
+        return false;
+    }
+
+    public function resolveUncaster(): Uncaster
+    {
+        return new $this->uncasterClass;
+    }
+}
--- a/src/Attributes/DefaultUncast.php
+++ b/src/Attributes/DefaultUncast.php
@@ -3,13 +3,12 @@
 namespace Spatie\DataTransferObject\Attributes;
 
 use Attribute;
-use ReflectionProperty;
-use ReflectionNamedType;
-use ReflectionUnionType;
 use JetBrains\PhpStorm\Immutable;
-use Spatie\DataTransferObject\Caster;
-use Spatie\DataTransferObject\Uncaster;
+use ReflectionNamedType;
+use ReflectionProperty;
+use ReflectionUnionType;
 use Spatie\DataTransferObject\Exceptions\InvalidUncasterClass;
+use Spatie\DataTransferObject\Uncaster;
 
 #[Attribute(Attribute::TARGET_CLASS)]
 class DefaultUncast
--- a/tests/Dummy/ComplexObjectUncaster.php
+++ b/tests/Dummy/ComplexObjectUncaster.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace Spatie\DataTransferObject\Tests\Dummy;
+
+use InvalidArgumentException;
+use Spatie\DataTransferObject\Uncaster;
+
+class ComplexObjectUncaster implements Uncaster
+{
+    /**
+     * @param array|mixed $value
+     *
+     * @return mixed
+     */
+    public function uncast($value): array
+    {
+        if (! $value instanceof ComplexObject) {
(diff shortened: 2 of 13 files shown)


brendt:
Looks nice, I remember a discussion about what `toArray` exactly should do in the past. There's an important distinction to make:

- representing a DTOs properties as an array, without making any assumptions about their values
- serializing a DTO to scalar values where nested DTOs are also uncasted

I'm not against the idea, but we need to make sure there's no confusion to end users about what does what.

morrislaptop:
Agree, this is the difference between `all()` and `toArray()` which already has a distinction in the readme?

> Note that all() will simply return all properties, while toArray() will cast nested DTOs to arrays as well.

We could implement I've just noticed on Laravel collections - `toJson()` which definitely removes confusion, but a string might be skipping a step which users might need.

brendt:
Agree that `toJson` is taking it a step too far. And thanks for reminding me about the README, I sometimes forget about stuff we wrote a while back 😅

The problem with making changes to `toArray` is that it's a breaking change, which means it'll take a while longer to actually be available in the package. Or am I missing something?

morrislaptop:
> The problem with making changes to toArray is that it's a breaking change, which means it'll take a while longer to actually be available in the package. Or am I missing something?

This idea is more of a feature which enables objects to be uncasted via `toArray` - this currently isn't possible. It would also be totally opt in via an explicit attribute, for example:

```
#[CastWith(ComplexObjectCaster::class)]
#[UncastWith(ComplexObjectCaster::class)]
public ComplexObject $complexObject;
```

```
class ComplexObjectCaster implements Caster, Uncaster
{
    // ...

    public function uncast(ComplexObject $value): array
    {
          return ['name' => $value->name];
    }
}
```

brendt:
Ooh, I get it! Yes, go for it!

src/Attributes/DefaultUncast.php
```diff
@@ -0,0 +1,57 @@
...
+#[Attribute(Attribute::TARGET_CLASS)]
+class DefaultUncast
+{
+    public function __construct(
+        #[Immutable]
+        private string $targetClass,
+        #[Immutable]
+        private string $uncasterClass,
+    ) {
+        if (! class_implements($this->uncasterClass, Uncaster::class)) {
```
morrislaptop, line 22: Noticed this check was not on the DefaultCast object - should we add it?

tests/Dummy/ComplexObjectUncaster.php
```diff
@@ -0,0 +1,25 @@
...
+class ComplexObjectUncaster implements Uncaster
+{
+    /**
+     * @param array|mixed $value
+     *
+     * @return mixed
+     */
+    public function uncast($value): array
+    {
+        if (! $value instanceof ComplexObject) {
```
morrislaptop, line 17: Can't provide a type hint since a child class can't be more specific in PHP. This is the best way to enforce it at run time and get typing.

morrislaptop:
@brendt had a first stab at this. It's very "copy paste" at the moment, there could be some refactoring to get the `resolveUncasterXXX` methods into the `DataTransferObjectProperty` class. Lmk if you think this is necessary or happy with it as is.

brendt:
Sorry for the delay in replying. I've been thinking some more about this. I think the best way to describe what we're trying to build here is "a modern way of serializing objects". While I there might be room for a PHP-8 focussed solution, I wonder whether this package is the place to do it. 🤔

aidan-casey:
@morrislaptop - I think your first comment "This library is fantastic for mapping incoming data to objects" is a good way of describing the focus of this package and perhaps its limitations on outgoing data are also intentional. As you know, this library was not built to handle outgoing data, but is a solution for transportation of data to other services within an application. These services could handle things such as: serializing the data for API response, formatting the data for PDF generation, etc.

Big additions, like serialization, seem like they take the focus away from good transportation of data and more toward a typed Laravel-esque model.

morrislaptop:
DTO's are objects that carry data between processes - I think that implies incoming data and outgoing data. The functionality seems so close to the casting that I think it makes sense here. 

This bidirectional approach is also seen with Laravel's casting and Symfony's serializer. Also, the `toArray` method is already leaning towards this direction?

It's particularly useful for sharing data types between projects / services. For example we can have a common type library with an `AccountResponse` data-transfer-object that's consumed between the quotes app and an accounts app. The accounts app returns the `AccountResponse` object and the quotes-app knows it can cast the response to the same `AccountResponse` object.

aidan-casey:
You may be right. I was more on the "don't implement casting" train when V3 was being planned as I honestly believe the logic involved takes away from what a DTO is. In other words, as [MSDN](https://docs.microsoft.com/en-us/previous-versions/msp-n-p/ff649585(v=pandp.10)?redirectedfrom=MSDN) put it "DTOs are simple objects that should not contain any business logic that would require testing."

All that to say, I am probably wrongly thinking of this from a different perspective than the direction this package has begun to take. [Martin Fowler](https://martinfowler.com/eaaCatalog/dataTransferObject.html) does seem to suggest the idea of some serialization logic within the DTOs as well.

brendt:
I'm currently undecided on this, as I see value in both points of view 😅

_if_ we decided to add this functionality, I wonder if `serialize` is a better word than `uncast`, as it's better know. 

That raises questions though: should be also implement PHP's `__serialize` and `__unserialize`? What about `json_encode` and `json_decode`?

morrislaptop:
Indeed a tricky one and I don't love the `uncast` naming, although it does show the inverse relationship to casting.

I think the library should remain "array based" and I worry that `serialize`, `json_encode` and `json_decode` could imply working with strings which goes too far.

Some other ideas:

**Normalizer::normalize() / NormalizeWith / DefaultNormalizeWith**

Naming taken from [Symfony Serializer](https://github.com/symfony/serializer/blob/5.x/Normalizer/NormalizerInterface.php) package. This is a good name but I'm not sure how well known this word is within the community.

**ArraySerializer::arraySerialize() / SerializeArrayWith / DefaultSerializeArrayWith**

Inspired by the [JsonSerializable](https://www.php.net/manual/en/class.jsonserializable.php) PHP interface which can return `mixed` to be used for json serialization. I like this one - well known words and clear on what it's doing :)

brendt:
So, I'm afraid it'll take a little more time to get this sorted out: my wife is going to give birth to our second son tomorrow, and after that I'll be a few weeks off the grid. We'll come back to this afterwards, or any of my colleagues can continue with it if they want to.

morrislaptop:
No problem and congratulations!

brendt:
I'm sorry for the very late reply!

I've been thinking about this one on and off for the past month or so, but I can't shake the feeling that this seems out of scope for this package. I might reiterate on that later, but I think I'll close it for now.

(I kind of feel bad for closing since it's clear you've put time and effort into it, so I hope you're not too disappointed…)

PR #189, opened 2021-04-10
Resolve default cast from parent classes

cappuc:
Calling `getAttributes` on a class returns only attributes applied to that class and not those applied to the parent classes, so we need to cycle through the entire hierarchy to load default casts from parent classes.


brendt:
Thanks!

brendt:
Tagged as 3.0.4

PR #191, opened 2021-04-10
Possibility to set default value

HanakJakub:
Previously it was possible to set default value directly in DTO class which I found really helpful so I make this small update to make it work also for this new version.


brendt:
Thanks! Will tag later today

brendt:
Tagged as 3.0.4

PR #196, opened 2021-04-16
Adds basic caching for reflection classes and properties.

### LINKED ISSUE #190: Cache reflections in v3
Hello,

First of all thank you very much for this amazing package. While navigating throw the source code of this library (inside the master branch [v3.x]) , I discovered we can improve the performance of casting objects by adding some cache system for the reflections used to fetch properties/classes metadata, for example inside the abstract class `DataTransferObject`, for each call to `all()` method you're performing:

```php
$class = new ReflectionClass(static::class);
$properties = $class->getProperties(ReflectionProperty::IS_PUBLIC);
...
```

Do you have some plans to integrate caching inside the v3 of this library? If yes, I can start working in a PR for this purpose.

aidan-casey:
This pull request adds support for "caching" of reflection classes and properties by storing them in static arrays keyed by the DTO FQDN.

This also modifies the `Spatie\DataTransferObject\DataTransferObject` class to use `Spatie\DataTransferObject\Reflection\DataTransferObjectClass` in the `all()` method as opposed to creating new reflection instances.

Closes #190

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to `data-transfer-object` will be documented in this file
 
+## Unreleased
+
+- Support caching of reflection classes and properties (#190)
+
 ## 3.0.4 - 2021-04-14
 
 - Support union types (#185)
--- a/composer.json
+++ b/composer.json
@@ -20,9 +20,9 @@
     },
     "require-dev": {
         "illuminate/collections": "^8.36",
+        "jetbrains/phpstorm-attributes": "^1.0",
         "larapack/dd": "^1.1",
-        "phpunit/phpunit": "^9.0",
-        "jetbrains/phpstorm-attributes": "^1.0"
+        "phpunit/phpunit": "^9.0"
     },
     "suggest": {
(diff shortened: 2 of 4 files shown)


brendt:
Have you done any tests on the performance impact on a couple of 1000 DTOs?

aidan-casey:
@brendt - Yes and the results actually impressed me on how efficient PHP 8 Reflection Classes are.

For 4,000 DTOs, this code produced a result in 0.6446430683136 seconds whereas the current package produced one in 0.72493410110474 seconds (consistently in that area).

For 10,000 DTOs, this code produced a result in 1.6914269924164 seconds whereas the current package produced one in 1.7899069786072.

So you're not gaining a whole lot to be honest. I ran those benchmarks using the traditional `microtime(true)` to capture start and end.

brendt:
In that case I don't think we'll add this functionality. It's a difference of 0.1 seconds in both the 4k and 10k example, and I want to avoid making the codebase more complex for those very small differences. Also keep in mind that memory consumption would go up with all those objects cached, so there's a downside to it as well.

Thanks for the effort though, it's good that we tried this out. I hope you're not too disappointed that your contribution won't be merged!

aidan-casey:
Not at all! I actually agree with your conclusion. 🙂

PR #198, opened 2021-04-22
Allow for DTO with optional parameters

eithed:
This PR will:
- allow for optional properties to stay optional and not be automatically set with null

---

Currently, if a property is optional, it will be automatically populated with null, even though the type indicates that it shouldn't be set on object at all. Consider following DTO:
```
class WithOptionalPropertyDto extends DataTransferObject
{
    public ?string $name;
}
```

It should be possible to construct this DTO per
```
$dto = new WithOptionalPropertyDto();
```
and
```
$dto = new WithOptionalPropertyDto([
   'name' => null
]);
```
but it doesn't mean that the DTOs are equal, as first one wasn't assigned a value and second one was. Currently, in both cases value is assigned and doing `$dto->toArray()` results in `['name' => null]`

### DIFF
--- a/src/DataTransferObject.php
+++ b/src/DataTransferObject.php
@@ -28,10 +28,10 @@ abstract class DataTransferObject
             if (array_key_exists($property->name, $args)) {
                 $property->setValue($args[$property->name]);
                 unset($args[$property->name]);
-            } else {
-                if ($property->hasDefaultValue()) {
-                    $property->setValue($property->getDefaultValue());
-                }
+            } elseif ($property->hasDefaultValue()) {
+                $property->setValue($property->getDefaultValue());
+            } elseif (!$property->allowsNull()) {
+                throw new \InvalidArgumentException("No value provided for required property {$property->name}");
             }
         }
 
--- a/src/DataTransferObject.php
+++ b/src/DataTransferObject.php
@@ -42,7 +42,7 @@ abstract class DataTransferObject
         $class->validate();
     }
 
-    public static function arrayOf(array $arrayOfParameters) : array
+    public static function arrayOf(array $arrayOfParameters): array
     {
         return array_map(
             fn (mixed $parameters) => new static($parameters),
@@ -50,7 +50,7 @@ abstract class DataTransferObject
         );
     }
 
-    public function all() : array
+    public function all(): array
     {
         $data = [];
 
@@ -73,7 +73,7 @@ abstract class DataTransferObject
         return $data;
     }
 
-    public function only(string ...$keys) : static
+    public function only(string ...$keys): static
     {
         $dataTransferObject = clone $this;
 
@@ -82,7 +82,7 @@ abstract class DataTransferObject
         return $dataTransferObject;
     }
 
-    public function except(string ...$keys) : static
+    public function except(string ...$keys): static
     {
         $dataTransferObject = clone $this;
 
@@ -91,12 +91,12 @@ abstract class DataTransferObject
         return $dataTransferObject;
     }
 
-    public function clone(...$args) : static
+    public function clone(...$args): static
     {
         return new static(...array_merge($this->toArray(), $args));
     }
 
-    public function toArray() : array
+    public function toArray(): array
     {
         if (count($this->onlyKeys)) {
             $array = Arr::only($this->all(), $this->onlyKeys);
@@ -109,7 +109,7 @@ abstract class DataTransferObject
         return $array;
     }
 
-    protected function parseArray(array $array) : array
+    protected function parseArray(array $array): array
     {
         foreach ($array as $key => $value) {
             if ($value instanceof DataTransferObject) {
--- a/src/DataTransferObject.php
+++ b/src/DataTransferObject.php
@@ -25,9 +25,14 @@ abstract class DataTransferObject
         $class = new DataTransferObjectClass($this);
 
         foreach ($class->getProperties() as $property) {
-            $property->setValue($args[$property->name] ?? $this->{$property->name} ?? null);
-
-            unset($args[$property->name]);
+            if (array_key_exists($property->name, $args)) {
+                $property->setValue($args[$property->name]);
+                unset($args[$property->name]);
+            } else {
+                if ($property->hasDefaultValue()) {
+                    $property->setValue($property->getDefaultValue());
+                }
+            }
         }
 
         if ($class->isStrict() && count($args)) {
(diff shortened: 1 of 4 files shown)


mfn:
But this also allows validly constructing:
```php
class StringDto extends DataTransferObject
{
    public string $name;
}

$dto = new StringDto([]);

$dto->toArray(); // #=> []

$dto->name; // 💥 
```
Effectively allowing all properties being uninitialized?

eithed:
> But this also allows validly constructing:
> 
> ```
> class StringDto extends DataTransferObject
> {
>     public string $name;
> }
> 
> $dto = new StringDto([]);
> 
> $dto->toArray(); // #=> []
> 
> $dto->name; // 💥 
> ```
> 
> Effectively allowing all properties being uninitialized?

Mhm - I've not noticed this case. I gather that the current behaviour of throwing Exception at the point of constructing DTO would still be wanted. I can proceed with handling this scenario if you think that presented use case makes sense.

src/DataTransferObject.php
```diff
@@ -25,9 +25,14 @@ public function __construct(...$args)
...
 
         foreach ($class->getProperties() as $property) {
-            $property->setValue($args[$property->name] ?? $this->{$property->name} ?? null);
-
-            unset($args[$property->name]);
+            if (array_key_exists($property->name, $args)) {
+                $property->setValue($args[$property->name]);
+                unset($args[$property->name]);
+            } elseif ($property->hasDefaultValue()) {
+                $property->setValue($property->getDefaultValue());
```
brendt, line 32: If the property has a default value already, why should we re-set it?

brendt (changes requested):

brendt:
One remark, but other than that this looks good to me :)

brendt:
Although I want to check in some of our projects first whether this might break anything.

brendt:
@rubenvanassche do you want to take a look at this, any problems you can think of that might come with it?

src/DataTransferObject.php
eithed: We don't necessarily need to set it here, but I inferred from previous behaviour that this was intended behaviour, per:

```
$property->setValue(... $this->{$property->name} ...);
```

I imagine without this line there would be a change in behaviour if the class is inherited and `setValue` is overloaded.

rubenvanassche:
I don't think we'll encounter problems with this PR 👍

brendt:
Just revisiting this. Why is setting it to `null` actually a problem?

```
public ?string $name;
```

Is there any case where you want to property to be uninitialized after constructing the object?

aidan-casey:
> Just revisiting this. Why is setting it to `null` actually a problem?
> 
> ```
> public ?string $name;
> ```
> 
> Is there any case where you want to property to be uninitialized after constructing the object?

I am curious of this too. It seems like a lot of effort for little gain, especially when you consider the fact that `null` values are considered _not set_ by PHP.

For further reference, [here](https://www.php.net/manual/en/language.types.null.php) are the official PHP comments on `null` values.

> The special null value represents a variable with no value. null is the only possible value of type null.
> 
> A variable is considered to be null if:
> 
> * it has been assigned the constant null.
> 
> * it has not been set to any value yet.
> 
> * it has been unset().

eithed:
I do agree that's it a bit of an effort. Theoretically, if using `it has not been set to any value yet.` from the reference, then yes, you are correct. One might ask then, what should be considered a valid `(new WithOptionalPropertyDto)->all()` - `[]` or `['name' => null]`?

From practical point of view, I'm using the DTO library in an ecosystem. There are two pain points I'm currently experiencing:
- DTO at one point is being returned as a resource and used in TS type - if it's being returned as `['name' => null]`, then either additional filtering needs to be done somewhere to strip null values or type cannot be defined as `type Foo = { name?: string }` but `type Foo = { name: string | null }`, which doesn't correspond to PHP type. Sure - filtering process can be done on output to handle this, but either way we're going back to `value null was set on object vs. value wasn't set on object`
- at one point I'm normalizing and denormalizing an object containing a DTO, using Symfony normalizers - if normalized object contains a DTO with optional property, then for some reason, during denormalization a `Cannot assign null to property` exception is thrown. Will try to come up with some reproducible code, but atm would safely ignore this one - there are many moving parts here and instantiating DTO with both `['someotherproperty' => 'bar']` and `['name' => null, 'someotherproperty' => 'bar']` should be possible.

aidan-casey:
@eithed - I like how this package has maintained it so far because it provides some standardization to `null` values and unset, but I can also see some value to aligning with how PHP does it. We are talking about a difference of a variable initialized with no value (`null`) vs. a property that has yet to be initialized (worth noting here that if a value is `unset()` it is cleared from memory and once again in an uninitialized state).

That said, I would still expect `all()` to return the property given the idea behind DTOs is to maintain the object structure and in the case of an API, I would probably return that key with the property set to null. This has the potential to be a breaking change since that array structure will change in both `all()` and `toArray()` and since the property is uninitialized, if you were to access it without setting a value it would now begin to throw an exception instead of returning `null`.

eithed:
> We are talking about a difference of a variable initialized with no value (`null`) vs. a property that has yet to be initialized (worth noting here that if a value is `unset()` it is cleared from memory and once again in an uninitialized state).

Yup, that's correct; I understand that :)

> That said, I would still expect all() to return the property given the idea behind DTOs is to maintain the object structure and in the case of an API, I would probably return that key with the property set to null. This has the potential to be a breaking change since that array structure will change in both all() and toArray() and since the property is uninitialized, if you were to access it without setting a value it would now begin to throw an exception instead of returning null.

While I don't agree with returning given property set to null I can agree with it being a breaking change - I don't think benefits outweight that danger. I'm happy to close and can revisit with my denormalization issue.

aidan-casey:
I'm not a maintainer, so @brendt will have to speak to this. I do see some value to tracking the uninitialized state, but I think the breaking change would technically require a semver increment to remain consistent.

Worth noting, this may be closely related to #188. It would still be breaking, but if uncasting were to be supported in the future, I could see the potential of just serializing the property as null if an uncaster is used like the example below.

```php
#[UncastWith(UninitializedPropertyCaster::class)]
public ?string $name;
```

brendt:
> what should be considered a valid (new WithOptionalPropertyDto)->all() - [] or ['name' => null]

My answer most definitely would be `['name' => null]`. That's because I think of a DTO's definition as the "source of truth", and not the data that's given to it.

I'm going to close this one, but I really appreciate the constructive discussion!

Issue #200, opened 2021-04-25
DefaultCast not repeatable


sboo:
When trying to use multiple default casts as mentioned in your documentation:
```
#[
    DefaultCast(DateTimeImmutable::class, DateTimeImmutableCaster::class),
    DefaultCast(MyEnum::class, EnumCaster::class),
]

```
I get an error that the attribute must not be repeated:

`PHP Error:  Attribute "Spatie\DataTransferObject\Attributes\DefaultCast" must not be repeated in /var/www/html/vendor/spatie/data-transfer-object/src/Reflection/DataTransferObjectProperty.php on line 143`

brendt:
Fixed in 3.1.1: https://github.com/spatie/data-transfer-object/releases/tag/3.1.1

PR #201, opened 2021-04-25
Makes DefaultCast repeatable

sboo:


sboo:
To solve the issue mentioned in #200

PR #204, opened 2021-04-30
Set boolean to class_implements

finexfloors:
Why class name?

https://www.php.net/manual/en/function.class-implements.php

The second parameter is boolean. I understand that there will be a cast, but still.

### DIFF
--- a/src/Attributes/CastWith.php
+++ b/src/Attributes/CastWith.php
@@ -15,7 +15,7 @@ class CastWith
         public string $casterClass,
         mixed ...$args
     ) {
-        if (! class_implements($this->casterClass, Caster::class)) {
+        if (! class_implements($this->casterClass, true)) {
             throw new InvalidCasterClass($this->casterClass);
         }
 
--- a/src/Attributes/CastWith.php
+++ b/src/Attributes/CastWith.php
@@ -15,7 +15,7 @@ class CastWith
         public string $casterClass,
         mixed ...$args
     ) {
-        if (! class_implements($this->casterClass, true)) {
+        if (! is_subclass_of($this->casterClass, Caster::class)) {
             throw new InvalidCasterClass($this->casterClass);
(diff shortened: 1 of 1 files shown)


src/Attributes/CastWith.php
```diff
@@ -15,7 +15,7 @@ public function __construct(
         public string $casterClass,
         mixed ...$args
     ) {
-        if (! class_implements($this->casterClass, Caster::class)) {
+        if (! class_implements($this->casterClass, true)) {
```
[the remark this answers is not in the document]
finexfloors: I meant that it will always be true here. A list of interfaces is also required here.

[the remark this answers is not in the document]
jdreesen: Yes, you are right. But I don't think the check makes sense at all. Would be strange to check if `$this->casterClass` just implements *something*. The intention was probably to check if it implements `Caster::class`.

brendt:
That's absolutely my bad; it should have been `is_subclass_of`. I'll fix this later this week.

src/Attributes/CastWith.php
```diff
@@ -15,7 +15,7 @@ public function __construct(
         public string $casterClass,
         mixed ...$args
     ) {
-        if (! class_implements($this->casterClass, Caster::class)) {
+        if (! is_subclass_of($this->casterClass, Caster::class)) {
```
jdreesen, line 18: I think we need the third parameter to be `true` here, because the first one is a string.
```suggestion
        if (! is_subclass_of($this->casterClass, Caster::class, true)) {
```

finexfloors: What for? There is true by default.

jdreesen: Oh, you're right!

brendt:
Thanks!

PR #206, opened 2021-05-03
Adds ArrayAccess casting to the ArrayCaster.

aidan-casey:
This PR adds the ability to cast objects that implement `ArrayAccess` to the `ArrayCaster`, it also adds the appropriate testing to ensure this works.

It should also be noted, I narrowed down the exception type to `LogicException` as opposed to `Exception` when casting an object other than an array or `ArrayAccess` is attempted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to `data-transfer-object` will be documented in this file
 
+## Unreleased
+
+- Add casting of objects that implement `ArrayAccess` to the `ArrayCaster`.
+
 ## 3.2.0 - 2021-04-30 
 
 - Support generic casters (#199)
--- a/README.md
+++ b/README.md
@@ -187,16 +187,37 @@ class ArrayCaster implements Caster
     ) {
     }
 
-    public function cast(mixed $value): array
+    public function cast(mixed $value): array|ArrayAccess
     {
-        if ($this->type !== 'array') {
-            throw new Exception("Can only cast arrays");
+        if ($this->type == 'array') {
+            return $this->castArray($value);
         }
 
+        if (is_subclass_of($this->type, ArrayAccess::class)) {
+            return $this->castArrayAccess($value);
+        }
+
+        throw new LogicException("Caster [ArrayCaster] may only be used to cast arrays or objects that implement ArrayAccess.");
+    }
+
+    private function castArray(mixed $value): array
+    {
         return array_map(
-            fn (array $data) => new $this->itemType(...$data),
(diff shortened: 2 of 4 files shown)


brendt:
Great, thanks!

swaner:
Noticed one thing about this when using it together with Eloquent Collection. Not sure why but when you initialise your DTO with a empty Collection the ArrayAccess caster still tries to create a new object. This will cause errors if not all properties are nullable.

This is my dto:
```
use App\Domain\ImportPayments\ImportTransactionDto;
class ImportPaymentDto extends DataTransferObject
{
    /** @var ImportTransactionDto[] */
    #[CastWith(ArrayCaster::class, itemType: ImportTransactionDto::class)]
    public Collection $transactions;
}
```

```
class ImportTransactionDto extends DataTransferObject
{
    public string $type;
}
```

And my initalization where I want a empty collection:
```
$activePaymentDto = new ImportPaymentDto([
            'transactions' => new Collection()
        ]);
```

This will throw an error:
TypeError: Cannot assign null to property App\Domain\ImportPayments\ImportTransactionDto::$type of type string in file /var/www/html/vendor/spatie/data-transfer-object/src/Reflection/DataTransferObjectProperty.php on line 46

#0 /var/www/html/vendor/spatie/data-transfer-object/src/Reflection/DataTransferObjectProperty.php(46): ReflectionProperty->setValue()
#1 /var/www/html/vendor/spatie/data-transfer-object/src/DataTransferObject.php(28): Spatie\DataTransferObject\Reflection\DataTransferObjectProperty->setValue()
#2 /var/www/html/vendor/spatie/data-transfer-object/src/Casters/ArrayCaster.php(52): Spatie\DataTransferObject\DataTransferObject->__construct()
#3 [internal function]: Spatie\DataTransferObject\Casters\ArrayCaster->Spatie\DataTransferObject\Casters\{closure}()
#4 /var/www/html/vendor/spatie/data-transfer-object/src/Casters/ArrayCaster.php(53): array_walk()


I managed to get rid of the error by checking count in the castArrayAccess method, but that requires the object to implement Countable, I am guessing there is a better way?
```
private function castArrayAccess(mixed $value): ArrayAccess
    {
        $arrayAccess = new $this->type();
        if (count($value) === 0) {
            return $arrayAccess;
        }
        
        array_walk(
            $value,
            fn (array $data) => $arrayAccess[] = new $this->itemType(...$data)
        );

        return $arrayAccess;
    }
    ```

aidan-casey:
@swaner this does seems odd. Thanks for reporting and including the detailed explanation!

@brendt I'll have a look at this today and see if I can replicate and include a fix.

PR #208, opened 2021-05-12
[Fix] Error : Unknown named parameter $var

lloricode:
when using array as same in v2 will return:
Error : Unknown named parameter $var

Edit:
see https://github.com/lloricode/paymaya-sdk-php/runs/2564106914?check_suite_focus=true

### DIFF
--- a/tests/ArrayCaster/ArrayV2CasterTest.php
+++ b/tests/ArrayCaster/ArrayV2CasterTest.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace Spatie\DataTransferObject\Tests\ArrayCaster;
+
+use Spatie\DataTransferObject\DataTransferObject;
+use Spatie\DataTransferObject\Tests\TestCase;
+
+class ArrayV2CasterTest extends TestCase
+{
+    /** @test */
+    public function version_2_will_casted()
+    {
+        $bar = new ParentDto(
+            [
+                'order_confirmation' => [
+                    'admin' => [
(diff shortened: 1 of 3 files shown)


brendt:
Can you provide a test and an example?

brendt:
(pressed the wrong button, didn't mean to close)

lloricode:
> Can you provide a test and an example?

yes i will :)

tests/ArrayCaster/ArrayV2CasterTest.php
```diff
@@ -0,0 +1,59 @@
...
+
+use Spatie\DataTransferObject\DataTransferObject;
+use Spatie\DataTransferObject\Tests\TestCase;
+
+class ArrayV2CasterTest extends TestCase
+{
+    /** @test */
+    public function version_2_will_casted()
+    {
+        $bar = new ParentDto(
```
lloricode, line 13: Hi @brendt,

this is our actual specific error when upgrading to v3.

Thank you

lloricode: on our project, its too difficult to refactor the DTO, since many using on this object

brendt:
If I understand you correct, you want to have interoperability between v2 and v3 objects?

lloricode:
> If I understand you correct, you want to have interoperability between v2 and v3 objects?

yes, we want v2 will completely work with v3 without any changes

brendt:
That's unfortunately not going to happen. We don't want to maintain interop between major versions.

PR #210, opened 2021-05-13
Support union type in casters

### LINKED ISSUE #209: Union type support
On caster class instantiation there's no check if object property has union type:
https://github.com/spatie/data-transfer-object/blob/d101e5185844eb92bf5fb8d0d1a200f1778e9802/src/Reflection/DataTransferObjectProperty.php#L85-L88
In that case it fails because ReflectionUnionType doesn't have getName() method.

daniser:
Closes #209

### DIFF
--- a/src/Reflection/DataTransferObjectProperty.php
+++ b/src/Reflection/DataTransferObjectProperty.php
@@ -82,11 +82,9 @@ class DataTransferObjectProperty
         /** @var \Spatie\DataTransferObject\Attributes\CastWith $attribute */
         $attribute = $attributes[0]->newInstance();
 
-        $typeNames = array_map(fn ($type) => $type->getName(), $this->extractTypes());
-
         return new $attribute->casterClass(
-            implode('|', $typeNames),
-            ...$attribute->args,
+            array_map(fn ($type) => $type->getName(), $this->extractTypes()),
+            ...$attribute->args
         );
     }
 
--- a/src/Reflection/DataTransferObjectProperty.php
+++ b/src/Reflection/DataTransferObjectProperty.php
@@ -82,27 +82,17 @@ class DataTransferObjectProperty
         /** @var \Spatie\DataTransferObject\Attributes\CastWith $attribute */
         $attribute = $attributes[0]->newInstance();
 
+        $typeNames = array_map(fn ($type) => $type->getName(), $this->extractTypes());
+
         return new $attribute->casterClass(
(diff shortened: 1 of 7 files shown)


src/Reflection/DataTransferObjectProperty.php
```diff
@@ -82,27 +82,17 @@ private function resolveCaster(): ?Caster
         /** @var \Spatie\DataTransferObject\Attributes\CastWith $attribute */
         $attribute = $attributes[0]->newInstance();
 
+        $typeNames = array_map(fn ($type) => $type->getName(), $this->extractTypes());
+
         return new $attribute->casterClass(
-            $this->reflectionProperty->getType()?->getName(),
+            implode('|', $typeNames),
```
brendt, line 88: How would this work without making changes to the `CastWith::__construct` attribute? It still expects a single class name. I understand the issue but I don't see how this PR fixes it.

brendt (changes requested):

daniser:
CastWith attribute doesn't need any changes AFAIK, but two supplied casters do. I've replaced "|"-delimited string to type name array and made needed changes to the casters.

src/Reflection/DataTransferObjectProperty.php
daniser: `CastWith` attribute doesn't need any changes AFAIK, but two supplied casters do. I've replaced "|"-delimited string with type name array and made needed changes to the casters.

brendt:
Can you submit tests that verify everything works as intended?

daniser:
> Can you submit tests that verify everything works as intended?

Done!

brendt:
Sorry for the late response.

Can you explain how this example works?

```php
    #[CastWith(RoundingCaster::class)]
    public float | int $bar;
```

`RoundingCaster` will always cast to `float`, so why would `int` still be needed there?

daniser:
> Sorry for the late response.
> 
> Can you explain how this example works?
> 
> ```
>     #[CastWith(RoundingCaster::class)]
>     public float | int $bar;
> ```
> 
> `RoundingCaster` will always cast to `float`, so why would `int` still be needed there?

It was my oversight; fixed.

brendt:
Thanks!

brendt:
Tagged: https://github.com/spatie/data-transfer-object/releases/tag/3.5.0

sidigi:
finaly

suyar:
Good job.

PR #213, opened 2021-05-18
Promoted dataTransferObject and validationErrors

tristankenney:
Hello! Just started using the library, and we're really enjoying it.

One use case that isn't easily supported (as far as I can see) is being able to access the results of a ValidationException – at present they are concatenated into a string.

This doesn't really work for our use case as we need to access individual messages and perform post-processing on the exceptions.

There's a number of ways that this could be done, but I thought the least invasive would be making the validationErrors available as a property (dataTransferObject has been promoted more for completeness).

### DIFF
--- a/src/Exceptions/ValidationException.php
+++ b/src/Exceptions/ValidationException.php
@@ -7,7 +7,7 @@ use Spatie\DataTransferObject\DataTransferObject;
 
 class ValidationException extends Exception
 {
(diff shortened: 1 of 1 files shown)


brendt:
Thanks!

brendt:
Tagged: https://github.com/spatie/data-transfer-object/commit/ad1547a1d35e81051520955c9f02f0eaafeb2ab9

PR #215, opened 2021-06-02
Migrate to PHP-CS-Fixer 3.x

shuvroroy:


.github/workflows/style.yml
```diff
@@ -15,7 +15,7 @@ jobs:
             - name: Run PHP CS Fixer
               uses: docker://oskarstark/php-cs-fixer-ga
               with:
-                  args: --config=.php_cs --allow-risky=yes
+                  args: --config=.php_cs.dist.php --allow-risky=yes
```
jdreesen, line 18: I think the new filename should be this (see: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/3.0/UPGRADE-v3.md#rename-of-files)

```suggestion
                  args: --config=.php-cs-fixer.dist.php --allow-risky=yes
```

brendt:
Nice, thanks!

PR #218, opened 2021-06-28
Fix ArrayAccess limitations

### LINKED ISSUE #216: ArrayAccess limitations
@swaner mentioned an issue with the `ArrayAccess` casting found in the `ArrayCaster` class in #206. I have a fix for that specific issue, but in implementing it I have discovered some oversights in my original implementation of the `ArrayAccess` caster which causes it to not work for very basic `ArrayAccess` implementations that do not also implement `IteratorAggregate`.

My current thinking is to implement some validation in the `ArrayCaster` that ensures the object also implements `IteratorAggregate`. I don't think this would really be a breaking change, as it would simply add in a check for a scenario we couldn't cover. As a bonus, we would also gain support for `ArrayObject` implementations in that caster.

@brendt - Any thoughts here? Happy to create a PR if this sounds like the right course.

aidan-casey:
This PR fixes #216 and the limitations it describes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to `data-transfer-object` will be documented in this file
 
+## Unreleased
+
+- Fix for an empty value being created when casting `ArrayAccess` objects (#216)
+- Add logic exception when attempting to cast `ArrayAccess` objects that are not traversable (#216)
+- Allow the `ArrayCaster` to retain values that are already instances of the `itemType` (#217)
+
 ## 3.3.0 - 2021-06-01
 
 - Expose DTO and validation error array in ValidationException (#213)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,7 @@ All notable changes to `data-transfer-object` will be documented in this file
 
 - Fix for an empty value being created when casting `ArrayAccess` objects (#216)
 - Add logic exception when attempting to cast `ArrayAccess` objects that are not traversable (#216)
-<<<<<<< HEAD
 - Allow the `ArrayCaster` to retain values that are already instances of the `itemType` (#217)
-=======
->>>>>>> 7e064acdd2afc392badb0aa348566fd91493fa7c
 
 ## 3.3.0 - 2021-06-01
 
--- a/README.md
+++ b/README.md
@@ -176,9 +176,11 @@ abstract class BaseDataTransferObject extends DataTransferObject
 
 ### Using custom caster arguments
 
-Any caster can be passed custom arguments, the built-in `ArrayCaster` [implementation](https://github.com/spatie/data-transfer-object/blob/master/src/Casters/ArrayCaster.php) is a good example of how this may be used.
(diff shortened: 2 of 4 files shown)


brendt:
Sorry it took so long, but this one looks good to merge! Can you rebase your branch? I've merged this one that's causing conflicts: https://github.com/spatie/data-transfer-object/pull/217

aidan-casey:
@brendt Hmm... So it looks like that PR isn't accounting for key / values (noticing I did not account for that in the array caster of this PR either). Looks like I'm going to need to do some refactoring here, so it will probably be Friday or Saturday before I can get that done.

brendt:
No problem :) Thanks for the heads-up!

aidan-casey:
@brendt - Would appreciate a review on this. A couple highlights:

- Renamed `makeItem` to `castItem` to fit with the rest of the class.
- Added `mapInto` method which casts and maps items to their appropriate key in an `array` or `Traversable` (this deduplicates code and keeps support for key mapping in arrays).
- Reverted type hinting on `$value` from `array` to `mixed` because setting it to `array` was a breaking change. I thought about union types here, but since the entry variable is `mixed` I just kept that.
- Removed contents of the `ArrayCaster` from the README and just added a link.

If this gets merged, I recommend squashing commits because I forgot about the style fixer and got in a fight with it. 😅

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ All notable changes to `data-transfer-object` will be documented in this file
 
 - Fix for an empty value being created when casting `ArrayAccess` objects (#216)
 - Add logic exception when attempting to cast `ArrayAccess` objects that are not traversable (#216)
+<<<<<<< HEAD
```
brendt: Careful, there's a merge conflict committed here

README.md
```diff
@@ -176,8 +176,11 @@ abstract class BaseDataTransferObject extends DataTransferObject
 
 ### Using custom caster arguments
 
-Any caster can be passed custom arguments, the built-in `ArrayCaster` implementation is such an example:
+Any caster can be passed custom arguments, the built-in `ArrayCaster` [implementation](https://github.com/spatie/data-transfer-object/blob/master/src/Casters/ArrayCaster.php) is a good example of how this may be used.
 
+<<<<<<< HEAD
```
brendt: Same here

CHANGELOG.md
aidan-casey: Yeah that was my bad (🤦🏻‍♂️), I caught it too late. Everything should be clean in the latest commit, so a squash and merge would clean up that history. Alternatively, I can clean up the Git history manually with a reset.

brendt: Don't worry about the history :)

Let me know when this is ready?

aidan-casey:
@brendt - I think this is good to go.

brendt:
Good!

brendt:
Sorry for the late response btw, going to tag it now.

brendt:
Tagged: https://github.com/spatie/data-transfer-object/releases/tag/3.4.0

PR #219, opened 2021-07-06
Add `fake` support

hello-liang-shan:
Hi, thanks for this great project, my english is not good, I will try my best to make sense

-----

```php
DTO::fake()
```

is very useful when testing, Im not sure if this is the best approach, but its nice to have this feature.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,10 +19,11 @@
         "php": "^8.0"
     },
     "require-dev": {
+        "fakerphp/faker": "^1.14",
         "illuminate/collections": "^8.36",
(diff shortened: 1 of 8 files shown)


brendt:
This is out of scope for this package. I suggest you read https://stitcher.io/blog/laravel-beyond-crud-09-test-factories — it's an older post but it still illustrates the point of how custom built test factories are the good solution here.

aidan-casey:
@hello-liang-shan - We have a package for this as well, if it is useful!

https://github.com/anteris-dev/data-transfer-object-factory

PR #220, opened 2021-07-07
Feat(toArray) Can transform keys to camel, studly and snake case

roke22:
Hello,
First of all thank you for your work. On some occasions I have found that the DTO that I use can be used to perform other actions if I convert my properties to camel case or snake case instead of having two different DTOs just by the name of the properties.

For them I have added 3 new methods which are toArraySnakeCase, toArrayStudlyCase and toArrayCamelCase

Its use would be:

```
$postData->toArraySnakeCase();
$postData->toArrayCamelCase();
$postData->toArrayStudlyCase(); 

```

This happen to us because our standard in the project is to name properties in camel case but some endpoint of 3rd party works with snake case, this try to be a solution when you want to use the same DTO to use with the method that use that 3rd party and don't break the rules of code.

Do you think it can be helpful?

Thank you.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,11 +1,11 @@
 {
-    "name": "spatie/data-transfer-object",
+    "name": "roke22/data-transfer-object",
     "description": "Data transfer objects with batteries included",
     "keywords": [
-        "spatie",
+        "roke22",
         "data-transfer-object"
     ],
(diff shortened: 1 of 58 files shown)


brendt:
We're not going to support that functionality. You can extend the base DTO class and add it on a project level.

PR #221, opened 2021-07-14
Feat(StrictTypes) Now works with strict types

roke22:
Hello,

Using this library I have run into a problem when trying to use strict types.
When creating a DTO with a Boolean property, if it receives a value greater than 1, it is set to true instead of giving a type error.

```
<?php
class MyDto extends DataTransferObject {
    public bool $field;
}

$dto = new MyDto(
    field:2
);

echo $dto->field;
```
The result is "true" as boolean but should be a error type if you are using strict_types=1.
I thought this was an error and I reported it to PHP and they replied stating everything was fine and that it was not an error, more info at https://bugs.php.net/bug.php?id=81258

So I have modified the library so that it can use strict types.
Now you can assign an attribute to indicate if the DTO uses strict types, by default it does not use them.

```
<?php

#[StrictTypes]
class MyDto extends DataTransferObject {
    public bool $field;
}

$dto = new MyDto(
    field:2
);
//This will fail with TypeError Exception
```
The most important change is in Src \ Reflection \ DataTransferObjectProperty.php and the tests for strict types in Tests \ StrictTypesTest.php

Review the modification since surely there is a more elegant way to do it.

I hope to be helpful.

### DIFF
--- a/src/Arr.php
+++ b/src/Arr.php
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types=1);
 
 namespace Spatie\DataTransferObject;
 
--- a/src/Attributes/CastWith.php
+++ b/src/Attributes/CastWith.php
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types=1);
 
 namespace Spatie\DataTransferObject\Attributes;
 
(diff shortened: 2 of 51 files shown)


brendt:
This was a conscious decision, and unfortunately you cannot change the package behaviour from outside, since that's how `declare(strict_types=1)` works. 

I'm not going to merge this one for now, since it'd possibly cause people's code to break. We can revisit it for the next major version.

PR #224, opened 2021-07-27
Support mapped properties

MaximeWillinger:
## Description

This pull request adds the ability to map DTO properties from source properties with different names.

This approach goes against the philosophy of doing a `1:1` mapping of DTOs, but it is undoubtedly helpful in the several cases that I present below.

### 1. Renaming a property when we have no control on the source data and/or for more clarity
```php 
class EventDTO extends DataTransferObject
{
    #[MapFrom('nbParticipants')]
    public string $participantCount;
}

$dto = new EventDTO([
    'nbParticipants' => 6,
]);
```

### 2. Mapping a nested property without using an additional DTO
```php 
class ProfileDTO extends DataTransferObject
{
    public string $name;

    #[MapFrom('address.city')]
    public string $city;
}

$dto = new ProfileDTO([
    'name' => 'John Doe',
    'address' => [
        'city' => 'London',
    ]
]);
```

### 3. Mapping properties from an indexed array
```php 
class UserDTO extends DataTransferObject
{
    #[MapFrom(0)]
    public string $firstName;

    #[MapFrom(1)]
    public string $lastName;
}

$dto = new UserDTO(['John', 'Doe']);
```

## Notes

This PR only requires the addition of a `resolveMappedProperty()` method in `DataTransferObjectProperty`.

However, handling the dot notation requires two additional array helpers (`Arr::get` and `Arr::accessible`). (_`Arr::get` has been simplified as default closure values are useless here._)

Despite the addition of these two methods, this PR is still lightweight but would significantly improve the DX for some cases without any significant constraint.

I'm open to change the name of the attribute if `MapFrom` doesn't suit you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,7 +292,7 @@ class NumberBetween implements Validator
 
 ## Mapping
 
-You can map a DTO property from an source property with a different name using the `#[MapFrom()]` attribute.
+You can map a DTO property from a source property with a different name using the `#[MapFrom]` attribute.
 
 It works with a "dot" notation property name or an index.
 
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ class MyDTO extends DataTransferObject
     #[NumberBetween(1, 100)]
     public int $a;
     
-    #[MapFrom('cityName')]
+    #[MapFrom('address.city')]
     public string $city;
 }
 ```
@@ -294,17 +294,23 @@ class NumberBetween implements Validator
 
 You can map a DTO property from an source property with a different name using the `#[MapFrom()]` attribute.
 
-It works with a property name or an index.
+It works with a "dot" notation property name or an index.
 
 ```php
 class PostDTO extends DataTransferObject
 {
     #[MapFrom('postTitle')]
     public string $title;
+    
+    #[MapFrom('user.name')]
+    public string $author;
 }
 
 $dto = new PostDTO([
     'postTitle' => 'Hello world',
+    'user' => [
+        'name' => 'John Doe'
+    ]
 ]);
 ```
 
--- a/README.md
+++ b/README.md
@@ -42,6 +42,9 @@ class MyDTO extends DataTransferObject
     
     #[NumberBetween(1, 100)]
     public int $a;
+    
+    #[MapFrom('cityName')]
+    public string $city;
 }
 ```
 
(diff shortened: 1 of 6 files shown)


src/Attributes/MapFrom.php
```diff
@@ -0,0 +1,16 @@
...
+
+use Attribute;
+use Spatie\DataTransferObject\Caster;
+use Spatie\DataTransferObject\Exceptions\InvalidCasterClass;
+
+#[Attribute(Attribute::TARGET_CLASS | Attribute::TARGET_PROPERTY)]
+class MapFrom
+{
+    public function __construct(
+        protected mixed $attribute,
```
abenerd, line 13: Maybe narrow this down, to be more explicit in which values are accepted?
```suggestion
        protected int | string $attribute,
```

README.md
```diff
@@ -287,6 +290,44 @@ class NumberBetween implements Validator
 }
 ```
 
+## Mapping
+
+You can map a DTO property from an source property with a different name using the `#[MapFrom()]` attribute.
```
innocenzi, line 295:
```suggestion
You can map a DTO property from a source property with a different name using the `#[MapFrom]` attribute.
```

src/Attributes/MapFrom.php
MaximeWillinger: Thanks for your feedback @abenerd! I just made the change.

src/Attributes/MapFrom.php
```diff
@@ -0,0 +1,16 @@
...
+
+use Attribute;
+use Spatie\DataTransferObject\Caster;
+use Spatie\DataTransferObject\Exceptions\InvalidCasterClass;
+
+#[Attribute(Attribute::TARGET_CLASS | Attribute::TARGET_PROPERTY)]
+class MapFrom
+{
+    public function __construct(
+        protected string | int $attribute,
```
brendt, line 13: Make this one public, so that you can use the attribute's instance (see the other comment)

src/Reflection/DataTransferObjectProperty.php
```diff
@@ -150,4 +151,15 @@ private function resolveCasterFromDefaults(): ?Caster
...
+
+    private function resolveMappedProperty(): string | int
+    {
+        $attributes = $this->reflectionProperty->getAttributes(MapFrom::class);
+
+        if (! count($attributes)) {
+            return $this->reflectionProperty->name;
+        }
+
+        return $attributes[0]->getArguments()[0];
```
brendt, line 163: use `$attributes[0]->newInstance()->name`

brendt (changes requested):
This is an awesome PR, thanks! One remark on attribute usage, but that's all.

src/Attributes/MapFrom.php
MaximeWillinger: Fixed in 6fd88a379b34965ea3c9ea687b84e4fc09a213c9.

src/Reflection/DataTransferObjectProperty.php
MaximeWillinger: Fixed in 6fd88a379b34965ea3c9ea687b84e4fc09a213c9.

brendt:
Thanks!

Tagged here: https://github.com/spatie/data-transfer-object/releases/tag/3.6.0

PR #225, opened 2021-08-12
Fix array assignment bug with strict dto's

marcoboers:
I came across an issue where the result of the custom `Spatie\DataTransferObject\Arr` helper is not used, which results in an exception with Strict DTO's. 
The custom `Spatie\DataTransferObject\Arr` helper is different from the Laravel `Arr` helper in that it returns an array instead of modifying an array passed by reference. 

This bug was introduced in 3.6.0.

### DIFF
--- a/src/DataTransferObject.php
+++ b/src/DataTransferObject.php
@@ -27,7 +27,7 @@ abstract class DataTransferObject
         foreach ($class->getProperties() as $property) {
             $property->setValue(Arr::get($args, $property->name) ?? $this->{$property->name} ?? null);
 
-            Arr::forget($args, $property->name);
+            $args = Arr::forget($args, $property->name);
         }
 
         if ($class->isStrict() && count($args)) {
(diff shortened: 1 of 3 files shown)


brendt:
Thanks!

brendt:
Tagged: https://github.com/spatie/data-transfer-object/releases/tag/3.6.1

milewski:
This PR now breaks my DTOs...

I had this:

```php
class CommitData extends DataTransferObject
{
    #[MapFrom('commit.committer.date')]
    public string $createdAt;

    public string $sha;
}
```

with this data now it throws an error:

```php
$data = [
    "sha" => "e201cfd551dc062b5f5bd80a85da8f142e3b3231",
    "commit" => [
        "committer" => [
            "date" => "2021-07-21T10:51:02Z",
        ],
    ],
]
```

```
Cannot assign null to property App\\DTOs\\CommitData::$sha of type string
```

brendt:
@MaximeWillinger you made #224, is it possible to fix both issues without reintroducing the bug that's fixed in this PR?

MaximeWillinger:
@brendt Of course! May I suggest streamlining the behaviors of DTO's `Arr` helpers with those of Laravel, to avoid confusion? This seems to solve #229 without breaking #225.

See my PR here: #230

Also @marcoboers, would you mind having a look?

PR #230, opened 2021-08-23
Streamline `Arr` helpers. Fix #229

### LINKED ISSUE #229: MapFrom bugs
**I can't use a root more than once.**

`user.email` returns null here
Screen Shot 2021-08-20 at 4 19 30 AM

If I switch up the order then `user.name` returns null
Screen Shot 2021-08-20 at 4 29 29 AM

**Sequence of properties matter?**

This works because `$description` is in the right order.
Screen Shot 2021-08-20 at 4 23 49 AM

This does not work because I moved it down?
Screen Shot 2021-08-20 at 4 31 34 AM

MaximeWillinger:


src/Arr.php
```diff
@@ -13,19 +13,23 @@ public static function only($array, $keys): array
...
     public static function except($array, $keys): array
     {
-        return static::forget($array, $keys);
+        static::forget($array, $keys);
+
+        return $array;
     }
 
-    public static function forget($array, $keys): array
+    public static function forget(&$array, $keys): void
```
brendt, line 21: Can we work around using references?

brendt:
Closed in favor of #231

PR #231, opened 2021-08-25
Correct behavior of Arr::forget with dot keys

### LINKED ISSUE #229: MapFrom bugs

shadowhand:
An alternative solution to #230 without using references.

Fixes #229

### DIFF
--- a/src/Arr.php
+++ b/src/Arr.php
@@ -32,19 +32,22 @@ class Arr
                 continue;
             }
 
-            $parts = explode('.', $key);
-
-            while (count($parts) > 1) {
-                $part = array_shift($parts);
-
-                if (isset($array[$part]) && is_array($array[$part])) {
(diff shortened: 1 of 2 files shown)


brendt:
Thanks!

brendt:
Released here: https://github.com/spatie/data-transfer-object/releases/tag/3.6.2

PR #236, opened 2021-09-06
Cast properties with `self` or `parent` type

sebdesign:
This PR fixes a bug when a type property has the `self` or `parent` keyword.

Here is a simple example with a nullable self property:
```php
{
    public string $name;
    public ?self $parent;
}

$dto = new MyDTO([
    'name' => 'a',
    'parent' => ['name' => 'b'],
]);

// TypeError: Cannot assign array to property MyDTO::$parent of type ?self
// vendor/spatie/data-transfer-object/src/Reflection/DataTransferObjectProperty.php:47
// vendor/spatie/data-transfer-object/src/DataTransferObject.php:29
```

The same applies for the `parent` keyword.

The reason is that the `DataTransferObjectProperty` cannot resolve the caster (e.g. `DataTransferObjectCaster`) for the property, because `resolveCasterFromType()` returns an empty array.
This happens because `$type->getName()` returns `self` so `class_exists('self')` is false.
In order to fix, this I'm parsing the type name and resolving the `self` and `parent` keywords to the concrete classes.

### DIFF
--- a/src/Reflection/DataTransferObjectProperty.php
+++ b/src/Reflection/DataTransferObjectProperty.php
@@ -7,6 +7,7 @@ use ReflectionAttribute;
 use ReflectionClass;
 use ReflectionNamedType;
 use ReflectionProperty;
+use ReflectionType;
 use ReflectionUnionType;
 use Spatie\DataTransferObject\Attributes\CastWith;
 use Spatie\DataTransferObject\Attributes\DefaultCast;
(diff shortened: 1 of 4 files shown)


brendt:
Looking good, thanks!

brendt:
Tagged: https://github.com/spatie/data-transfer-object/releases/tag/3.7.1

PR #237, opened 2021-09-12
Ensure value is an array before trying to destruct

milewski:
Currently the following throws an error:

```php
class Email extends DataTransferObject
{
    public string $value;
    
    public function __construct($value)
    {
        parent::__construct([ 'value' => $value ]);
    }
}
```

```php
$works = new UserDTO([
    'email' => new Email('email@example.com'),
]);

$fail = new UserDTO([
    'email' => 'email@example.com',
]);
```

```
TypeError : Only arrays and Traversables can be unpacked
```

This PR fixes it.

### DIFF
--- a/src/Arr.php
+++ b/src/Arr.php
@@ -95,4 +95,13 @@ class Arr
 
         return array_key_exists($key, $array);
     }
+
+    public static function wrap($value): array
+    {
+        if (is_null($value)) {
+            return [];
+        }
+
+        return is_array($value) ? $value : [$value];
+    }
 }
--- a/src/Casters/DataTransferObjectCaster.php
+++ b/src/Casters/DataTransferObjectCaster.php
@@ -2,6 +2,7 @@
 
 namespace Spatie\DataTransferObject\Casters;
 
+use Spatie\DataTransferObject\Arr;
 use Spatie\DataTransferObject\Caster;
 use Spatie\DataTransferObject\DataTransferObject;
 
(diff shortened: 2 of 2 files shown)


brendt:
If you want to override the constructor, you'll need to manually take care of passing the right data.

Furthermore: it's highly discouraged to override the constructor, you should use static constructors, we'll make the constructor private in the next major release.

milewski:
I think you misunderstood the issue highlighted on this PR, the issue here is not with the constructor etc it was a quickly dirty example.. the issue is if you have a DTO like this:

```php

// Assuming I have this
class UserDTO extends DataTransferObject
{
    public Email $email;
}

// I want to be able to construct it either ways:

new UserDTO(email: new Email('email@email.com')); // Passing an instance to it
new UserDTO(email: 'email@email.com'); // Or letting DataTransferObject make one instance of it
```
Currently, it is impossible... because even if the Email object knows how to be constructed from an Array or A string as param it will crash because the DTO is trying to always unpack... and under no circumstance unpacking anything other than an Array is allowed...

https://github.com/spatie/data-transfer-object/blob/7960781ad8eb69e58e7915dcb1d8a912c35f3671/src/Casters/DataTransferObjectCaster.php#L15-L23

So All this PR does is ensure it is always unpacking what can be unpacked.....

brendt:
Gotcha. I don't think your PR covers all bases though. Simply wrapping a value when it's not an array doesn't seem like a foolproof method. I'd like to see some tests at first

milewski:
Okay, can then this be re-opened and set as a draft then? later I will try to write some tests for it

PR #239, opened 2021-09-14
Read `#[Strict]` also from parent classes

### LINKED ISSUE #238: Setting `#[Strict]` on parent is not applied on children
```php
#[Strict]
abstract class DtoBase {}

final class Dto extends DtoBase {}
```

then `Dto` is not strict because `DataTransferObjectClass::isStrict()` reads attributes only for current instance but not its parents. 

Is it intended or a bug? Thanks!

simPod:
Fixes #238

### DIFF
--- a/src/Reflection/DataTransferObjectClass.php
+++ b/src/Reflection/DataTransferObjectClass.php
@@ -2,6 +2,7 @@
 
 namespace Spatie\DataTransferObject\Reflection;
 
+use ReflectionAttribute;
 use ReflectionClass;
 use ReflectionProperty;
 use Spatie\DataTransferObject\Attributes\Strict;
@@ -66,6 +67,19 @@ class DataTransferObjectClass
 
     public function isStrict(): bool
     {
-        return $this->isStrict ??= ! empty($this->reflectionClass->getAttributes(Strict::class));
+        if (!isset($this->isStrict)) {
+            $attribute = null;
+
+            $reflectionClass = $this->reflectionClass;
+            while ($attribute === null && $reflectionClass !== false) {
+                $attribute = $reflectionClass->getAttributes(Strict::class)[0] ?? null;
+
+                $reflectionClass = $reflectionClass->getParentClass();
+            }
+
+            $this->isStrict = $attribute !== null;
+        }
+
+        return $this->isStrict;
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to `data-transfer-object` will be documented in this file
 
+## Unreleased
+
+- `#[Strict]` is passed down the inheritance chain so children are strict when parent is strict (#239)
+
 ## 3.7.1 - 2021-09-09
 
 - Cast properties with self or parent type (#236)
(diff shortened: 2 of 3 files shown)


src/Reflection/DataTransferObjectClass.php
```diff
@@ -66,6 +67,22 @@ public function validate(): void
...
+
+    /**
+     * @param array<ReflectionAttribute> $attributes
+     *
+     * @return array<ReflectionAttribute>
+     */
+    private function getStrictAttributesWithParent(ReflectionClass $reflectionClass, array $attributes = []): array
+    {
+        $attributes = array_merge($attributes, $reflectionClass->getAttributes(Strict::class));
+        $parent = $reflectionClass->getParentClass();
```
brendt, line 81: This should be in a loop because there could several levels of inheritance. I've done something like this in the past:

```php
$attribute = null;

while ($attribute === null && $reflectionClass) {
    $attribute = $reflectionClass->getAttributes(Strict::class)[0] ?? null;

    $reflectionClass = $reflectionClass->getParent();
}
```

brendt: I haven't tried this out, but it's something along those lines.

simPod: I have used recursion which does the same, see passing test. Do you prefer a loop?

brendt: Yes, a loop is clearer than recursive functions.

src/Reflection/DataTransferObjectClass.php
```diff
@@ -66,6 +67,15 @@ public function validate(): void
...
+        $attribute = null;
+
+        $reflectionClass = $this->reflectionClass;
+        while ($attribute === null && $reflectionClass !== false) {
+            $attribute = $reflectionClass->getAttributes(Strict::class)[0] ?? null;
+
+            $reflectionClass = $reflectionClass->getParentClass();
+        }
+
+        return $this->isStrict ??= $attribute !== null;
```
jdreesen, line 79: This now isn't caching correctly anymore.

simPod: Can you elaborate? Tests pass.

jdreesen: Previously, the value of `$this->isStrict` was only calculated when it wasn't set (i.e. not initialized). Now it's calculated every time (actually the `??=` is now quite useless, if not harmful, as `$this->isStrict` is never updated with the re-calculated value once it's set).

simPod: oh, state, right... good catch

jdreesen: A solution would be to extract the actual calculation into a private method and call it like `return $this->isStrict ??= $this->determineIsStrict();`

Or to wrap everything into a check if `$this->isStrict` is set and only do the calculation if it is not (and not use the `??=` operator anymore).

brendt:
Thanks!

brendt:
Released: https://github.com/spatie/data-transfer-object/releases/tag/3.7.2

PR #242, opened 2021-09-30
V4

brendt:
WIP


brendt:
This PR already includes the change to the private constructor. I've got some doubts:

- Having to use `call_user_func_array("{$this->itemType}::new", $data);` is more complex, but this is not a deal breaker
- You can use anonymous classes in tests anymore, I fixed this by adding an `AnonymousDataTransferObject` in the tests namespace in order to prevent having to make dozens of custom DTO classes
- Going back to the original goal of making the constructor private (allowing several ways of constructing a DTO, eg: with and without validation), I wonder whether it's worth the massive breaking change? Is a feature that will be useful to enough people?

brendt:
@aidan-casey feel free to share your opinion.

src/Casters/ArrayCaster.php
```diff
@@ -60,7 +60,7 @@ private function castItem(mixed $data)
         }
 
         if (is_array($data)) {
-            return new $this->itemType(...$data);
+            return call_user_func_array("{$this->itemType}::new", $data);
```
aidan-casey, line 63: We may have to check whether the item type is a Data Transfer Object, and then use the `new` method. That said, we may also want to use `newWithoutValidation` and then have to iterate through traversables to validate the DTOs.

src/DataTransferObject.php
```diff
@@ -17,14 +17,27 @@ abstract class DataTransferObject
...
+        return $dataTransferObject
+            ->setUp($class, ...$args)
+            ->validate($class);
+    }
+
+    private function __construct()
+    {
+    }
+
+    protected function setUp(DataTransferObjectClass $class, ...$args): static
```
aidan-casey, line 35: This could potentially be:

```php
public static function newWithoutValidation(...$arguments): static
{
    //
}
```

src/DataTransferObject.php
```diff
@@ -17,14 +17,27 @@ abstract class DataTransferObject
 
     protected array $onlyKeys = [];
 
-    public function __construct(...$args)
+    public static function new(...$args): static
```
aidan-casey, line 20: This could potentially just be:

```php
public static function new(...$arguments): static
{
    return static::newWithoutValidation($arguments)->validate();
}
```

src/DataTransferObject.php
```diff
@@ -35,13 +48,20 @@ public function __construct(...$args)
             throw UnknownProperties::new(static::class, array_keys($args));
         }
 
+        return $this;
+    }
+
+    protected function validate(DataTransferObjectClass $class): static
```
aidan-casey, line 54: Do you think there may be value to making the validate method public? For example, the new method would set the properties on the DTO, and then the validate method would be called.

It's probably a good idea to put the validation logic in its own resolver or something as well.

```php
public function validate(): static
{
    //
}
```

aidan-casey: On this note, if we have validation providers, all the logic for validating DTOs could be handled by a `DataTransferObjectValidationProvider` that is attributed to the base class (just like the caster).

For example:

```php
#[ValidateWith(DataTransferObjectValidationProvider::class)]
abstract class DataTransferObject
{
    //
}
```

aidan-casey:
> Having to use `call_user_func_array("{$this->itemType}::new", $data);` is more complex, but this is not a deal breaker

This could be changed to `return $this->itemType::new($data)` and would be auto-completed if we add an instance check on the `itemType` property.

> You can(not) use anonymous classes in tests anymore, I fixed this by adding an `AnonymousDataTransferObject` in the tests namespace in order to prevent having to make dozens of custom DTO classes

We can always add a helper method for this, though I think my personal preference would be to see a few well-thought-out stubs created.

> Going back to the original goal of making the constructor private (allowing several ways of constructing a DTO, eg: with and without validation), I wonder whether it's worth the massive breaking change? Is a feature that will be useful to enough people?

I think it's worthwhile more for assisting in simplifying the underlying code while allowing for things like rolling up validation exceptions.

src/DataTransferObject.php
```diff
@@ -7,23 +7,72 @@
...
+    }
+
+    private static function describe(): ClassDescriptor
+    {
+        if (! isset(static::$description)) {
+            static::$description = new ClassDescriptor(static::class);
+        }
+
+        return static::$description;
+    }
```
aidan-casey, line 62: This method won't work in the long run. We probably need to leave the `__construct()` method empty, have a property value resolver, and make this method non-static.

src/DataTransferObject.php
brendt: > I think we will need to make the validate method public for it to work with sub-DTOs that need to be validated

Can't you call protected methods from sub DTOs? I thought that was possible but haven't tested it.

brendt: About the validation provider: can you refresh my memory on when it would be useful to override it? What I'm questioning here is whether it needs to be as configurable as you describe?

src/DataTransferObject.php
brendt: Absolutely, feel free to refactor if you're working in the code.

src/DataTransferObject.php
brendt: Can you explain why it won't work?

src/Descriptors/PropertyDescriptor.php
```diff
@@ -0,0 +1,65 @@
...
+
+    public function getReflection(): ReflectionProperty
+    {
+        return $this->property;
+    }
+
+    /**
+     * @return array<ReflectionType>
+     */
+    public function getTypes(): array
```
brendt, line 30: We could use spatie/better-types here, it's one of the use cases I had in mind

src/Resolvers/CastResolver.php
```diff
@@ -0,0 +1,44 @@
...
+                $value
+            );
+        }
+
+        return $arguments;
+    }
+
+    protected function castArgument(PropertyDescriptor $property, mixed $value): mixed
+    {
+        $attributes = $property->getReflection()->getAttributes(CastWith::class);
```
brendt, line 30: We can also use spatie/better-types here. 

I'm also fine with being able to resolve the cast information from without the property descriptor to keep this code a little more readble.

eg `$property->getCastWith` or something alike. 

I think our custom attributes are part of the "schema" or "description" of the property, it's only what we do with it that should be handled by this class. Let me know your thoughts though!

src/Resolvers/MapFromResolver.php
```diff
@@ -0,0 +1,43 @@
...
+        foreach ($this->descriptor->getProperties() as $property) {
+            $arguments = $this->mapArgument($arguments, $property);
+        }
+
+        return $arguments;
+    }
+
+    private function mapArgument(array $arguments, PropertyDescriptor $property): array
+    {
+        $attributes = $property->getReflection()->getAttributes(MapFrom::class);
```
brendt, line 28: Same notes here: let's use spatie/better-types and discuss whether we move the attribute resolving part to the descriptor instead.

src/DataTransferObject.php
aidan-casey: > Absolutely, feel free to refactor if you're working in the code.

Will do!

src/DataTransferObject.php
aidan-casey: > Can't you call protected methods from sub DTOs? I thought that was possible but haven't tested it.

By "sub DTOs" I am referencing data transfer object properties that are cast using `newWithoutValidation` and then need to be validated during the validation step (so as to roll up all the errors into a single exception).

I am envisioning a call like:

```php
PersonData::newWithoutValidation(firstName: 'Aidan', lastName: 'Casey')->validate();
```

> About the validation provider: can you refresh my memory on when it would be useful to override it? What I'm questioning here is whether it needs to be as configurable as you describe?

Right now rules can be created at a property level, but it would be awesome if a provider could handle a whole class as well. This provider would be passed an instance of the `ClassDescriptor` which would allow it to intimately validate the properties. For example, someone (a.k.a. - me  👀 ) could create a LaravelValidationProvider which uses attributes as rules, but the provider handles reading those attributes and making assertions based on them.

This functionality could be utilized for validating the sub DTOs as well by checking that the property type extends a DTO and calling the `validate` method on it as well. Similar to the `DataTransferObjectCaster` but for validation.

src/DataTransferObject.php
aidan-casey: In the following code:

```php
class Dto1 extends DataTransferObject
{
    public string $property;
}

class Dto2 extends DataTransferObject
{
    public string $otherProperty;
}

$dto1 = new Dto1(property: 'test');

$dto2 = new Dto2(otherProperty: 'some other test');
```

Both `$dto1::description` and `$dto2::description` equal the `ClassDescriptor` created with `Dto2::class` because it is static.

src/Resolvers/CastResolver.php
aidan-casey: Agreed. I think it might be better to add a `getAttribute` method (utilizing `spatie/better-types`) so that a handy helper for other Attributes is present as well.

aidan-casey:
Overhauled a lot this evening, be curious about your thoughts, @brendt.

PR #246, opened 2021-10-09
Test against php 8.1

sergiy-petrov:


sergiy-petrov:
@aidan-casey hey, thanks for the answer. What about merging it into V3 branch?

suyar:
> I think our PHP 8.1 feature development should be focused only in v4, but if making v3 compatible doesn't take too much work, then why not? Since we're requiring `"php": "^8.0"`, we're already saying we'll support 8.1.
> 
> So I'll merge this for now, we'll see how 8.1 development evolves in the coming weeks.

Is there a release plan for V4 version?

brendt:
Nothing concrete right now.

Issue #249, opened 2021-10-26
Cannot assign null to property of type bool


augustusnaz:
Creating the below object attempts to set the boolean properties to null.

```php
final class Address extends DataTransferObject
{
    ...

    /**
     * Address country name or code
     *
     * @var string
     */
    public string $country;

    /**
     * If the address is public
     *
     * @var bool
     */
    #[MapFrom('is_public')]
    #[MapTo('is_public')]
    public bool $isPublic = false;
}
```

```php
$address = new Address(
    country: "AU",
    // is_public: false,
);
```

I suspect it's caused [here](https://github.com/spatie/data-transfer-object/blob/0b7dcb67cb85958c5bec42edea749092c9522d91/src/DataTransferObject.php#L29).

Can anyone reproduce this?

sergyu91ma:
any workaround until fix is released?

stevebauman:
Dang, auto-closed this too eh

PR #251, opened 2021-10-27
Fixing Bug #249

gerpo:
Default values work with MapFrom.

The issue with assign null to type, was that the default value was ignored if the mapping was not present. 
Now it will fall back to the default if the mapped property is not present.


stevebauman:
This should be re-opened 👍

aidan-casey:
@gerpo - Can you modify this PR to use the default parameter for `Arr::get()` as opposed to the null coalescing operator? I'll release this patch once that is done.

gerpo:
> @gerpo - Can you modify this PR to use the default parameter for `Arr::get()` as opposed to the null coalescing operator? I'll release this patch once that is done.

Done :)

aidan-casey:
Thank you, @gerpo. Apologies for the delay!

joecampo:
This is awesome. I literally just ran into this today. Talk about good timing. Thanks @aidan-casey & @gerpo you're lifesavers. ❤️

sebdesign:
I think this is introducing a new bug, or if it's intended, that's a breaking change.

Before this PR, properties that had a default value, would be assigned the default value if the argument was `null`.
This was because of the null coalescing operator.
If `Arr::get($args, $property->name)` was `null`, then `$this->{$property->name}` would return the default value.

Now with `Arr::get($args, $property->name, $property->getDefaultValue())`, If the `$args` array has a key named `$property->name`, it will return its value whether it's `null` or not. The `$property->getDefaultValue()` will only be returned when the `$args` array does **NOT** have this key.

The following test was green until this PR.

```php
$data = [
    'title' => 'Hello world',
    'description' => null,
];

$dto = new class ($data) extends DataTransferObject {
    public string $title;

    public string $description = 'Test Text';
};

$this->assertEquals('Hello world', $dto->title);
$this->assertEquals('Test Text', $dto->description);
```

Now this test throws an error:
```
TypeError: Cannot assign null to property Spatie\DataTransferObject\DataTransferObject@anonymous::$description of type string

/vendor/spatie/data-transfer-object/src/Reflection/DataTransferObjectProperty.php:48
/vendor/spatie/data-transfer-object/src/DataTransferObject.php:29

**Comment by aidan-casey** [comment]
@sebdesign - Please see #282. The behavior you are describing was never intended, documented, or a supported feature. It was a bug that was fixed in this PR.

**Comment by pr4xx** [comment]
For me this introduced a new bug in my application. I have a base class which extends `DataTransferObject`. In the constructor, just before calling `parent::__construct();`, I set some default values which I cannot set just using the class properties (for example object instantiation). Prior to this update this worked:
```php
class BaseDTO extends DataTransferObject
{
    public MyObject $myObject;
    
    public function __construct(...$args)
    {
        // set default
        $this->myObject = new MyObject();
        // call parent
        parent::__construct(...$args);
    }
}
```
Now this fails because when the parent constructor runs, it tries to set `$myObject` to `null` because it has no default value. Prior to this update it would have used the existing instance created from me.
What would be the correct way to assign complex default values prior to assigning data from json/array?

PR #253, opened 2021-11-05
Feature/cast properties from method

SimonJnsson:
This PR proposes the addition of casting properties on a DTO without using attributes and instead specifying a castAttribute method on the DTO. This is useful for one-off casts, that would otherwise require a seperate Caster class.

Example usage:

```php
{
    public string $name;
    
    public function castName($value)
    {
        return strtoupper($value);
    }
}
```

In this example, the DTO would automatically find the castName function and use this alongside any other provided Casters.

Any input on the behaviour of using this in combination with a CastWith attribute would be greatly appreciated.

Consider the following example, where ToLowerCaster is a simple caster that casts the value to all lower case.

```php
{
    #[CastWith(ToLowerCaster::class)]
    public string $name;
    
    public function castName($value)
    {
        return strtoupper($value);
    }
}
```
What should the expected result be in such a case?

For now, it simple uses the existing logic to cast via any CastWith attribute, and then passes the casted value to the cast method, meaning the current implementation would result in name being uppercased.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,22 @@ class ComplexObjectWithCast
 }
 ```
 
+### Caster methods
+
+Instead of creating a custom caster and using `CastWith`, you can create a cast method on the DTO, using the naming convention `cast<PropertyName>`.
+
+```php
+class MyDTO extends DataTransferObject
+{
+    public string $name;
+    
+    public function castName($value)
+    {
+        return strtoupper($value);
+    }
+}
+```
+
 ### Default casters
 
 It's possible to define default casters on a DTO class itself. These casters will be used whenever a property with a given type is encountered within the DTO class.
--- a/src/Reflection/DataTransferObjectProperty.php
+++ b/src/Reflection/DataTransferObjectProperty.php
@@ -5,6 +5,7 @@ namespace Spatie\DataTransferObject\Reflection;
 use JetBrains\PhpStorm\Immutable;
 use ReflectionAttribute;
 use ReflectionClass;
+use ReflectionMethod;
 use ReflectionNamedType;
 use ReflectionProperty;
 use ReflectionType;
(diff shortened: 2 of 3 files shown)


aidan-casey:
I've wondered about this one myself from time to time, and I can definitely see the use case with those one-off situations. 

My personal conclusion in the past has been that as nice as it would be, that logic probably doesn't belong in a DTO and the more public methods present, the more muddled an object that should be fairly simple becomes.

I could maybe have my mind changed on that with a protected method, but that would then just make the implementation more complex.

At the end of the day, it's a call for @brendt, but those are just a few of my thoughts.

SimonJnsson:
> I've wondered about this one myself from time to time, and I can definitely see the use case with those one-off situations.
> 
> My personal conclusion in the past has been that as nice as it would be, that logic probably doesn't belong in a DTO and the more public methods present, the more muddled an object that should be fairly simple becomes.
> 
> I could _maybe_ have my mind changed on that with a protected method, but that would then just make the implementation more complex.
> 
> At the end of the day, it's a call for @brendt, but those are just a few of my thoughts.
> 
> **Edit**: Should this be implemented, it may be better found in V4, where a resolver could first be run for the attribute casting, and then a second one follow it for method casting.

I did consider that the public method would slightly crowd the developer experience, but was unsure if it would be possible to apply the same solution to protected methods.

brendt:
I like the effort put into this, but I'm not going to merge it. Not that it's not good, but I'm simply hesitant to introduce more ways of doing to same thing:

- It complicates the mental overhead for users
- It adds more paths to reason about in our code, which — in the end — we have to maintain.

So really appreciate you thinking along, but for now the answer is no :)

booni3:
I would love to know how you would deal with a situation, where the casting relies on multiple fields to determine logic. This is less of casting from one value but more of a logic path to determine the required value. To illustrate, this is how I am currently doing it... but there must be a cleaner way inline with the package?

```php
class MyOrderDTO extends BaseOrderDTO
{
    public function __construct(...$args)
    {
        $order = $args[0];
        $order['status'] = $this->status($order);

        parent::__construct($order);
    }

...

    protected function fillStatus(array $order): string
    {
        if(in_array($order['status'], ['new', 'open'])){
            if($order['filledSize'] > 0 && $order['remainingSize'] > 0){
                return 'part-filled';
            }

            return 'open';
        }

        return 'closed';
    }
}
```

aidan-casey:
@booni3 - This really isn't a casting problem here. It's more about the logic that defines what state the order is in. Since a data transfer object simply _transfers the data_, that falls out of the scope of what a DTO should be doing.

So really that status should be an enum, state machine, or some other type that keeps track of the state and determined by whatever is initializing the DTO.

booni3:
@aidan-casey Thanks for the input. I maybe thought so much. In my case, I am using the DTO's for transforming API calls. i.e.

```php
MyOrderDTO::fromApiResponse($response)
```

And this works really well. I was using the previous Spatie package on PHP7 and you could add all of this transformation logic quite cleanly. In the case of the API data, we are normalizing the data but as fields between providers are not common we sometimes need to use some logic to infer things such as status.

In PHP8, everything else is now so much cleaner than doing it the way I felt wrong.

aidan-casey:
@booni3 - Makes sense! You might want to look at something like [Symfony's Serializer](https://github.com/symfony/serializer) to handle some of the transformation. 🙂

PR #263, opened 2021-12-10
PHP 8.1 Intersection types support

### LINKED ISSUE #262: PHP 8.1: Intersection types support
Currently [intersection types](https://wiki.php.net/rfc/pure-intersection-types) are not supported by the package.

It is not possible to construct on object of the following type at the moment
```php
class SomeDto extends DataTransferObject
{
    public Foo & Bar $foobar;
}
```

zjbarg:
Fixes #262

### DIFF
--- a/tests/IntersectionDtoTest81.php
+++ b/tests/IntersectionDtoTest81.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace Spatie\DataTransferObject\Tests;
+
+use Spatie\DataTransferObject\Attributes\DefaultCast;
+use Spatie\DataTransferObject\Caster;
+use Spatie\DataTransferObject\DataTransferObject;
+
+/**
+ * @requires PHP >= 8.1
+ */
+class IntersectionDtoTest81 extends TestCase
+{
+    /** @test */
+    public function intersection_types_are_allowed()
+    {
+        $value = new SomeFooBar;
+
+        $dto = new DtoWithIntersectionTypeProp(foobar: $value);
+
+        $this->assertEquals($value, $dto->foobar);
+    }
+
+    /** @test */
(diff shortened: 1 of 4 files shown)


tests/IntersectionDtoTest81.php
```diff
@@ -0,0 +1,54 @@
...
+
+class SomeFooAndSomeBarCaster implements Caster
+{
+    public function cast(mixed $value): SomeFooBar
+    {
+        return new SomeFooBar;
+    }
+}
+
+#[DefaultCast(SomeFoo::class.'&'.SomeBar::class, SomeFooAndSomeBarCaster::class)]
```
zjbarg, line 50: This whole thing with the default cast on intersection type is not making since to me. Probably should be ignored.

aidan-casey:
Hey @zjbarg,

First of all, thanks for the PR! I appreciate the work you have put into this.

Personally I think I would be inclined to wait on supporting this until V4. There are some changes that could make this much more clean to implement. 

What do you think, @brendt?

zjbarg:
hey @aidan-casey,

Thanks for taking the time.

If you ignore the stuff regarding DefaultCast, the change is acutally only one line. 

```php
class DataTransferObjectProperty
{
    // ...
    private function extractTypes(): array
    {
        // ...
        return match ($type::class) {
            ReflectionNamedType::class => [$type],
            ReflectionUnionType::class => $type->getTypes(),
            ReflectionIntersectionType::class => $type->getTypes(), // ++++++++
        };
    }
}
```

aidan-casey:
@zjbarg walk me through this so I can understand a bit better the goal. The only thing this changes is the behavior of casting. So in the scenario that I have `ArrayAccess&Traversable` it's simply going to attempt casting the first one it can find a caster for, not try to cast both.

I think to really support this we would need to check for a caster and validate that the value meets the intersection criteria.

PR #264, opened 2021-12-30
Stop suggesting phpstan/phpstan

ymilin:
This suggest section was added in #135, But all functionalities were removed in tag 3.0.0


mfn (approved):
Yes, makes sense; I'm the author of that PR all of this became obsolete with the major versions following the v1 release.

PR #267, opened 2022-02-03
Add onlyFilled method

yusufonur:
The onlyFilled method ensures that only filled values are returned.

For example, when we transfer the values that come with FormRequest to DTO, nullable fields come as null and may cause an update in the database.

Example:

    public function update(UserUpdateRequest $request, User $user)
    {
        $userData = (new UserData($request->validated()))->onlyFilled();

        $user = $this->userService->update($userData, $user);

        return new UserResource($user);
    }

If date_of_birth does not appear in UserUpdateRequest and this field is not mandatory in UserData, this field comes as null and the relevant field is updated as null in the table. We can avoid this problem because we get non-null fields with the onlyFilled method.

### DIFF
--- a/src/DataTransferObject.php
+++ b/src/DataTransferObject.php
@@ -86,6 +86,19 @@ abstract class DataTransferObject
         return $dataTransferObject;
     }
 
+    public function onlyFilled(): static
+    {
+        $dataTransferObject = clone $this;
+
+        foreach ($dataTransferObject->toArray() as $property => $value) {
+            if (isset($value)) {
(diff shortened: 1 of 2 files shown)


brendt:
What happens if you _want_ to save `null` as a value in de database?

yusufonur:
> What happens if you _want_ to save `null` as a value in de database?

In cases with this condition, we can continue with ->all() without using onlyFilled. This is how I do it and I think it has been very useful so far.

brendt:
I don't think we want to support this on the package level. You're always free to make your own DTO implementations on a project-level, extending from our base class to add additional functionality.

PR #268, opened 2022-02-15
Transform DTO with DataTransformer

coldic3:
I find it very helpful to transform the entire DTO during initialization and achieve ready to use DTO. `DataTransformer` can be used in `#[TransformWith]` attribute.

### DIFF
--- a/src/Attributes/TransformWith.php
+++ b/src/Attributes/TransformWith.php
@@ -1,0 +1,17 @@
+<?php
+
+namespace Spatie\DataTransferObject\Attributes;
+
+use Attribute;
+use Spatie\DataTransferObject\DataTransformer;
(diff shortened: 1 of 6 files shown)


brendt:
Can you give an example of what this PR does?

brendt:
Why not add a static constructor on the DTO itself?

coldic3:
I didn't think about it TBH. For sure it's much more simple but still see some advantages of a transformer. It's another responsibility and it's good to have it in a separate object, transformers are reusable, can transform only part of args and let mappers do the rest.

brendt:
From my side, it's quite a lot of extra code to maintain, while I think static constructors are a perfectly fine solution. So I won't merge this in for now.

PR #277, opened 2022-05-23
Added enum caster

Elnadrion:
Hi there!
Me and my team often use enum caster, in our DTOs and it would be gr8 to have it by default in package
The only problem that I faced - I can't cover this caster with tests inside the package, cuz package supports PHP 8.0, and enums comes since 8.1

Here is usage example:

image

P.S. I will also add this information in readme, if you accept this PR

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
     "require-dev": {
         "illuminate/collections": "^8.36",
         "larapack/dd": "^1.1",
-        "phpunit/phpunit": "^9.0",
+        "phpunit/phpunit": "^9.5.5",
         "jetbrains/phpstorm-attributes": "^1.0"
     },
     "autoload": {
--- a/src/Casters/EnumCaster.php
+++ b/src/Casters/EnumCaster.php
@@ -2,24 +2,18 @@
 
 namespace Spatie\DataTransferObject\Casters;
 
-use LogicException;
 use Spatie\DataTransferObject\Caster;
 
 class EnumCaster implements Caster
 {
     public function __construct(
         private array $types,
-        private string $enumType,
-        private string $enumValueType
+        private string $enumType
     ) {
     }
 
     public function cast(mixed $value): mixed
     {
-        if (gettype($value) !== $this->enumValueType) {
-            throw new LogicException("$this->enumType can only be casted from $this->enumValueType");
-        }
-
         return $this->enumType::from($value);
     }
 }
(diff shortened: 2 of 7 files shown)


Elnadrion:
P.P.S. Also, I believe that type checking with "enumValueType" variable is quite doubtful and we just can use default php error... What do you think?

P.P.P.S. Also I've noticed the v4 branch (after I created this PR) with PHP 8.1 only. I can try to merge this PR into v4 branch and cover this feature by tests

danilopolani:
This would be a great addition on v4, I think it's not possible in v3 because of the PHP version.
I agree that just using the backed enum `::from()` method is enough, without the `enumValueType`.

Thank you for the PR, hopefully we'll have this soon 🤞

Elnadrion:
Hello :)

I got rid of extra type checking (this PR became so lightweight 😁)

I believe we actually can put this PR into v3 branch. It just won't work without PHP 8.1. Also I'm going to add this functionality in v4 in another PR, with all needed tests, blackjack and hookers 😂

Elnadrion:
I've made v4 of this feature in this PR https://github.com/spatie/data-transfer-object/pull/279

Btw @danilopolani I didn't make generic annotations. Tbh - I don't use them by my self and don't sure about syntax too

aidan-casey:
@Elnadrion - A couple requests before this PR is merged:

1. This PR will fail for basic enumerations. I think we would be better off checking to ensure the enum exists and is backed. If it does not, throw a helpful `LogicException`.
2. Can you use the `tryFrom` method on the backed enum and if it fails, throw a helpful `LogicException` as well?
3. Can you add tests to the PR?

Elnadrion:
@aidan-casey 
As I mentioned above - I dunno how I can cover this code with tests with PHP 8.0 support

aidan-casey:
@Elnadrion - PHPUnit has a `@requires` annotation. See [here](https://phpunit.readthedocs.io/en/9.5/incomplete-and-skipped-tests.html#skipping-tests-using-requires) for more information.

aidan-casey:
@Elnadrion - Almost there I think. I couple of notes:

1. The `prefer-lowest` tests are failing due to a bug in PHPUnit. Go ahead and bump the dependency version on that to `^9.5.5` to resolve.
2. Can you update the exception messages to be more similar to existing ones? I am thinking something like:
    - `Caster [EnumCaster] may only be used to cast backed enums. Received [$this->enumType].`
    - `Couldn't cast enum [$this->enumType] with value [$value].`
3. Can you update the subclass check to use the string `'BackedEnum'` as opposed to importing the class? Just feel more comfortable with that in PHP 8 land.

Elnadrion:
@aidan-casey done it 👌

aidan-casey:
@Elnadrion thanks!

PR #278, opened 2022-06-02
Ability to get only filled fields from DTO

Elnadrion:
Hi again :) 

Sometimes we need to get only those values that we set when we create DTO (for example: if we need update our model from $dto->toArray()). This PR will provide us this functionality


aidan-casey:
@Elnadrion - Thanks for the PR! See #267 for more information about why I am closing this. Depending on how v4 shapes up we may consider some functionality along these lines then.

PR #280, opened 2022-06-03
Enable object with MapTo properties to be cloned (Fix #272)

### LINKED ISSUE #272: Unexpected behavior on clone and MapTo
Hey,

I want to report an unexpected behavior/ bug when using clone and `MapTo`. 

Here is an example: 
```php
class ExampleData extends DataTransferObject
{
    #[MapTo('account_id')]
    public ?int $account;

    #[CastWith(CarbonImmutableCaster::class), MapTo('expires_at')]
    public ?CarbonImmutable $expiresAt;
}
```
Reproduce the problem:
```php
$o = new ExampleData(
  account: 1,
  expiresAt: '2020-01-01 12:00:00'
);

$nO = $o->clone(account: 2); // I would expect $nO to have account 2 with same expiresAt

$nO->toArray() // [account: 2, expires_at: null]

```

This happens because of how `clone`/ `toArray` works: 
```php 
    public function clone(...$args): static
    {
        return new static(...array_merge($this->toArray(), $args));
    }
```
The code above does something like: 
```php
return new static(...array_merge([
  'account_id' => 1,
  'expires_at': '2020-01-01 12:00:00'
], [
  'account' => 2
]);
```

And this is where the problem comes in. The property name is `expiresAt` so `expires_at` is not matched and instead dropped. Only `account` works, as it is specified correctly. 

This maybe could be avoided with a `MapFrom` I guess, but I expect it to work without.

Could we potentially change/fix this behavior?

gerpo:
This PR adds a flag to ignore the MapTo attribute. 
In some cases, it could be useful to get the original keys instead of the mapped ones when  you export a data object to an array.
This is also the easiest way to fix the bug in issue #272 as we can just pass true for the $keepOrginalKeys flag and don't walk into any array merging issues. 

With the default value of false we also don't risk any breaking changes. If the flag is not passed, we get the same behavior as right now.

### DIFF
--- a/src/DataTransferObject.php
+++ b/src/DataTransferObject.php
@@ -60,7 +60,7 @@ abstract class DataTransferObject
             }
 
             $mapToAttribute = $property->getAttributes(MapTo::class);
-            $name = count($mapToAttribute) && !$keepOriginalKeys ? $mapToAttribute[0]->newInstance()->name : $property->getName();
+            $name = count($mapToAttribute) && ! $keepOriginalKeys ? $mapToAttribute[0]->newInstance()->name : $property->getName();
 
             $data[$name] = $property->getValue($this);
         }
--- a/src/DataTransferObject.php
+++ b/src/DataTransferObject.php
@@ -46,7 +46,7 @@ abstract class DataTransferObject
         );
     }
 
-    public function all(): array
+    public function all(bool $keepOriginalKeys = false): array
     {
         $data = [];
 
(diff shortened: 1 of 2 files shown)


aidan-casey:
Thanks for the PR, @gerpo. What's the use case for this outside of fixing #272?

danielhomer:
Could this be merged in? This is blocking us from cloning a DTO with multiple `MapTo` props.

freekmurze:

PR #287, opened 2022-06-29
Feature: Add implicit caster and NullableCaster

danilopolani:
The PR adds an `ImplicitCaster` interface that, similar to `ImplicitRule` on Laravel, allows to run casters on "null" values too.
It also ships a new caster, `NullableCaster`, that allows to set a default fallback on null values and proves how `ImplicitCaster` works and what it can do.

Thank you in advance! 🤞


aidan-casey:
Hey @danilopolani, thank you for your efforts here. I will be holding off on merging this for the time being as, in general, data transfer objects should have minimal logic and purely be concerned with transferring the data it contains. This logic adds one-offs to the existing functionality, and would be better off being handled at the instantiation of the object.

danilopolani:
> @danilopolani - Don't get me wrong, I do see your point, it just doesn't fit with the long term vision and has the potential to add some maintenance overhead as we look to shift to where this package needs to go.
> 
> Adding casters to begin with _was_ a mistake. The long term plan is that such functionality is sent through a deserialization / denormalization layer before being actually passed to the object. This will provide much more flexibility than continuing to compound if statements in the package. 🙂

Oh, make sense, thank you :)

aidan-casey:
No problem! Again, I do appreciate the work you put into this, it just doesn't fit with the vision at this time.

PR #289, opened 2022-07-12
Add simple boolean caster

kemp:
This PR adds a boolean caster with tests included.

The array of "truthy" booleans is sourced from Laravel's boolean validation: https://github.com/laravel/framework/blob/904bf5630a95ecf1d6623eb502ba7b751875faf4/src/Illuminate/Validation/Concerns/ValidatesAttributes.php#L422

Thanks!

### DIFF
--- a/src/Casters/BooleanCaster.php
+++ b/src/Casters/BooleanCaster.php
@@ -1,0 +1,15 @@
+<?php
+
+namespace Spatie\DataTransferObject\Casters;
+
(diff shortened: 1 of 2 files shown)


aidan-casey:
@kemp - Thanks for the PR! Unfortunately I think we are going to wait on merging this at this time. Adding default casters to begin with did not really fit with the vision for this package, and we need to continue to work away from that.

PR #290, opened 2022-07-13
Fallback to default value if json value is null

chady:
https://github.com/spatie/data-transfer-object/pull/251 broke an old behavior that I was depending on.

The change uses `Arr::get` which returns the default value only if the key does not exist in the array. But if the key exists *and* is `null`, then it will try to set the property to `null` instead of the expected default value.


chady:
@aidan-casey I came back to argue that this should be a supported feature, but while formulating my argument I realized why the current implementation is the correct approach 😄 

Thank you for the reply and sorry for taking up your time.

PR #291, opened 2022-07-29
Support nested arrays using "." notation in MapTo attributes

morloderex:
Hello Spatie.

## Use case
I am currently using this package to map over an external api response. This external api's uses a lot of nested array properties which I want to flatten down in the DTO properties.

## The problem
The mapping part is working great on nested arrays when using the `MapFrom` attribute but i find it a little bit strange that using the `MapTo` attribute doesn't take into account nested data structures using `.` notation in the same way as `MapFrom` currently does. 

Given this DTO definition:
```php
class DTO extends DataTransferObject {
    #[MapTo('count.testing')]
    #[MapFrom('count.testing')]
    public int $originalCount;

    #[MapTo('hero.villain')]
    #[MapFrom('hero.villain')]
    public string $villain;
};

new DTO(['count' => ['testing' => 'testing'], 'hero' => ['villain' => 'testing']]);
```

Currently when calling the `all` method on the DTO generates an array like this:
```php
[
    "count.testing" => 'testing',
    "hero.villain" => 'testing'
]
```
 But i expected it to be:

 ```php
[
    "count" => ['testing' => 'testing'],
    "hero" => ['villain', => 'testing'],
]
 ```
 
 This is what this pr is trying to change so we can effectively align the use of `MapFrom` attribute and `MapTo` attributes with each other.
 
 I am aware that this is probably a huge breaking change so it would probably require a major update.

### DIFF
--- a/src/Arr.php
+++ b/src/Arr.php
@@ -95,4 +95,44 @@ class Arr
 
         return array_key_exists($key, $array);
     }
+
+    /**
+     * Set an array item to a given value using "dot" notation.
+     *
+     * If no key is given to the method, the entire array will be replaced.
+     *
+     * @param  array  $array
+     * @param  string|int|null  $key
+     * @param  mixed  $value
+     * @return array
+     */
+    public static function set(&$array, $key, $value)
+    {
+        if (is_null($key)) {
+            return $array = $value;
+        }
+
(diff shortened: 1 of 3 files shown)


freekmurze:

PR #293, opened 2022-08-03
Adding support for strict type checks

YamenSarah:
We are utilizing dtos to enforce our API signatures.
So we have a preference to not allow for type coercion. Unfortunately since this is using setValue, our strict_types declarations are ignored.

This PR adds support for strict type checks utilizing the existing Strict attribute following the same rules enforced by native strict type declarations.

I saw an older issue about this that said it would be revisited, but haven't seen anything else related.
Let us know if there is any alternative

### DIFF
--- a/README.md
+++ b/README.md
@@ -335,6 +335,12 @@ new StrictDto(
 );
 ```
 
+### Type Coercion
+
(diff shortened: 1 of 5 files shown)


YamenSarah:
@brendt bump :)

freekmurze:

PR #299, opened 2022-08-29
keep unknow properties for tracking even in non strict DTO s

crux:
I'm unsure about the naming, but something like this would be pretty helpful. I would like to use strict DTOs to check and enforce my APIs, but then, the receiving code would break whenever a malformed DTO arrives. So I switch to non strict DTOs, but then, all messages arriving which contain some unspecified junk would go unnoticed.

The idea here with the PR is to keep that unknown attribute names for reporting purpose available. So I could log and warn about them in my stack, without actually bailing out on any bad message.

### DIFF
--- a/src/DataTransferObject.php
+++ b/src/DataTransferObject.php
@@ -17,6 +17,8 @@ abstract class DataTransferObject
 
     protected array $onlyKeys = [];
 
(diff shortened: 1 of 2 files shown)


freekmurze:

PR #300, opened 2022-09-13
Ignoring uninitialized object properties

m1nyasha:
This PR was created to address an issue with non-passable values in DTOs.

In the current implementation, properties whose values are not passed to the class constructor are set to their default values.

```php
<?php
use Spatie\DataTransferObject\DataTransferObject;

class UpdateArticleDTO extends DataTransferObject
{
    public ?string $title;
    public ?string $body;
}

$data = ['title' => 'Some title'];

(new UpdateArticleDTO($data))->toArray(); // ['title' => 'Some title', 'body' => null]
```

I suggest adding the ability to simply ignore these properties on getting by adding the `OnlyInitialized` attribute.

```php
<?php
use Spatie\DataTransferObject\Attributes\OnlyInitialized;
use Spatie\DataTransferObject\DataTransferObject;

#[OnlyInitialized]
class UpdateArticleDTO extends DataTransferObject
{
    public ?string $title;
    public ?string $body;
}

$data = ['title' => 'Some title'];

(new UpdateArticleDTO($data))->toArray(); // ['title' => 'Some title']
```

This can be useful for API requests to edit records in a database.

### DIFF
--- a/src/Attributes/OnlyInitialized.php
+++ b/src/Attributes/OnlyInitialized.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Spatie\DataTransferObject\Attributes;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_CLASS)]
+class OnlyInitialized
+{
+
+}
--- a/src/DataTransferObject.php
+++ b/src/DataTransferObject.php
@@ -26,6 +26,10 @@ abstract class DataTransferObject
         $class = new DataTransferObjectClass($this);
 
         foreach ($class->getProperties() as $property) {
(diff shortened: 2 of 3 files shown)


aidan-casey:
@m1nyasha - Thanks for taking the time to put together this PR! Unfortunately, I will be holding off from merging this functionality at this time.

Data transfer objects provide a consistent definition for what data the developer can expect to successfully pull from them. Allowing uninitialized properties breaks this pattern by requiring the developer to check whether those values are set, instead of being able to simply grab that property with a consistent value (e.g., it will always be `string` or `null`).

Thanks again for opening the PR!

PR #301, opened 2022-09-14
allow enum casting not only enum value casting

regnerisch:
This PR allows to cast enums, not just enum values.


aidan-casey:
Thanks, @regnerisch! I think this is a good approach even for backed enums as it will now allow the value to be an enum instance. 👍🏼

regnerisch:
That's the idea. I was a little irritated I have to do something like `$data = MyData(origin: MyEnum::SOMETHING->value)` to use the enum caster, so I created this PR :)

PR #305, opened 2022-09-19
✨ Allow intersection types

juliomotol:
This PR aims to add support for Intersection type introduced in PHP8.1


freekmurze:

PR #315, opened 2022-10-31
Add instanceof check to EnumCaster

Bloemendaal:
A fix for when the instance of the enum itself is returned. The caster currently expects a string or int, but it's kinda weird that it doesn't except the enum itself.

image


Bloemendaal:
nvm, our package was not up-to-date

